PR #1, opened 2018-06-06
PMOS, PNP & NPN symbols

eqvinox:
just the symbols for now, no devices... one step at a time...


ubruhin:
Thanks :)

The PMOS looks good for me. But the BJTs miss labels for name/value. And could you rotate the BJTs so they have the same rotation as mosfets?

ubruhin:
Oops, I completely forgot this pull request, sorry! :scream: 

In the meantime these symbols were already added to this library, so this PR no longer needed ;)

PR #2, opened 2018-07-24
Add SOT-23 N-Channel MOSFET

dbrgn:
Add Toshiba SSM3K329R N-Channel logic-level MOSFET with SOT23-3 footprint.


ubruhin:
Please use zero-rotation as defined in IPC:

grafik

And why is in the default footprint pad 3 larger than the other two pads? I would use the same pad size for all pads...

Otherwise it looks good :+1:

dbrgn:
> And why is in the default footprint pad 3 larger than the other two pads? I would use the same pad size for all pads...

That's according to the SOT-23 datasheet by NXP: https://www.nxp.com/docs/en/package-information/SOT23.pdf I used the wave soldering footprint and made it a bit smaller. Should I make the pad the same size as the other two pads?

dbrgn:
Fixed. Good now?

dbrgn:
I also tried to align the pads to the grid.

Pad 3 is centered on the 2.54mm grid. Pad 1 and 2 don't fit on the 2.54mm grid unfortunately, but at least they fit on the 1.27mm grid.

img

ubruhin:
> That's according to the SOT-23 datasheet by NXP: https://www.nxp.com/docs/en/package-information/SOT23.pdf I used the wave soldering footprint and made it a bit smaller. Should I make the pad the same size as the other two pads?

Ah, but the wave soldering footprint shouldn't be the default footprint, nor should it be called "default".

I would suggest to just rename "default" to "wave soldering" and change the order of the footprints so that "reflow" is at top (the footprint at top is automatically used as default).

> Pad 3 is centered on the 2.54mm grid. Pad 1 and 2 don't fit on the 2.54mm grid unfortunately, but at least they fit on the 1.27mm grid.

Oh, I guess I misunderstood you in the librepcb chat ;) I meant that for *Symbols* the origin should be near to the center and aligned to grid. But for footprints we cannot decide on which grid to place the pads - they must always be placed *exactly* according datasheet. In the board editor we have to solve the grid snap issue anyway because there will always exist footprints with off-the-grid pads.

Regarding the origin of footprints: I'm not sure if IPC defines where exactly the origin should be. If it does, do it that way. If not, I would suggest to place it (more or less) in the center of the *mechanical body*, not in the mathematical center of all pad coordinates. This way it is more intuitive to drag&drop footprints because you can grab them where the body is, and because the pick&place export needs the coordinate of the body center which would then always be [0,0].

dbrgn:
> Ah, but the wave soldering footprint shouldn't be the default footprint, nor should it be called "default".

That's why I shrunk the pads, to make them suitable for hand soldering :slightly_smiling_face: (The wave soldering pads are even bigger.)

> If you like, add one more footprint which is the same as the reflow footprint but with larger pads (but still all pads same size) for hand soldering and set this as default footprint.

I can do that.

> they must always be placed exactly according datasheet

Huh, does the datasheet specify the origin?

ubruhin:
> That's why I shrunk the pads, to make them suitable for hand soldering slightly_smiling_face (The wave soldering pads are even bigger.)

Ah, I see :)

> Huh, does the datasheet specify the origin?

I mean the distance between the pads is specified in the datasheet, so we mustn't align pad distances to a multiple of e.g. 2.54mm. Because of this, we can't align (all) pads to grid.

And yes, origin may be specified in datasheet because it's relevant for pick&place.

We should copy this documentation: http://kicad-pcb.org/libraries/klc/F6.2/ :smiley:

ubruhin:
@dbrgn what's the state of this PR? The package and its category are now already contained in the library, so only the device would need to be added.

dbrgn:
Yes, I'll have to update this sometime.

dbrgn:
@ubruhin branch updated.

I also added a P-channel MOSFET.

Btw, is it correct to put those MOSFETs in the base library? I wonder if they should go into a Toshiba library instead.

ubruhin:
> Btw, is it correct to put those MOSFETs in the base library? I wonder if they should go into a Toshiba library instead.

Hmm yeah probably it would make sense to move them into a Toshiba library...

The devices look good IMO, although I didn't check the pinout ;)

dbrgn:
Closing: https://github.com/LibrePCB-Libraries/Toshiba.lplib/pull/1

PR #3, opened 2018-11-01
Add sym/cmp for pushbuttons

dbrgn:
Open by default:

screenshot

Closed by default:

screenshot

Open questions:

- Should the invisible grab area be put there or not?


ubruhin:
- Improve names, e.g. "Momentary switch, open" -> "Push Button Momentary NO" and "Momentary switch, closed" -> "Push Button Momentary NC"
  - Title Case
  - I think the symbol you have drawn is only used for push buttons, not for any kind of switches
  - Starting the names with "Push Button" makes them easier to find in sorted lists
  - "NO" and "NC" (normally open/closed) are terms commonly used for switches/pushbuttons
- Grab area is reasonable, but needs some changes:
  - Doesn't cover the "button" of the normally-open symbol -> increase height
  - Should not overlap with pins -> decrease width
  - Must be filled, otherwise only the area of its lines can be grabbed
  - As the lines aren't required to grab, better change line width to zero
- Maybe place {{VALUE}} at the bottom of the symbol?
- Set component default value to "{{PARTNUMBER or DEVICE}}"

dbrgn:
> Improve names, e.g. "Momentary switch, open" -> "Push Button Momentary NO" and "Momentary switch, closed" -> "Push Button Momentary NC" 

A momentary switch is not necessarily a push button. A push button is a momentary switch though.

> Starting the names with "Push Button" makes them easier to find in sorted lists

The push button devices would start with "Push Button".

> I think the symbol you have drawn is only used for push buttons, not for any kind of switches

According to [Wikipedia](https://en.wikipedia.org/wiki/Electronic_symbol#Switches) it seems to be the IEEE symbol for momentary switches :slightly_smiling_face: 

> "NO" and "NC" (normally open/closed) are terms commonly used for switches/pushbuttons

Ah, good point! I updated the names (including title casing)

> Grab area is reasonable, but needs some changes: 
> Maybe place {{VALUE}} at the bottom of the symbol?

Done:

no

nc

> Set component default value to "{{PARTNUMBER or DEVICE}}"

Done.

ubruhin:
> According to Wikipedia it seems to be the IEEE symbol for momentary switches

But in the IEEE standard this symbol is called "Pushbutton":

grafik

For other momentary switches, I guess these symbols at the right apply:

grafik

There are many different kinds of switches (pushbutton/rotary/slider/dipswitch/etc., most of them available as momentary and locked variant). All of them have different symbols (see for example [here](https://www.electrical-symbols.com/electric-electronic-symbols/switch-symbols.htm)). I wonder if we should create separate components for all these types, or only a few "generic" switch components. I think at this time I would go with separate components...

dbrgn:
You're right :slightly_smiling_face: I'll adjust the naming and will try to fix the name on Wikipedia too.

dbrgn:
Done. I used the name "Pushbutton" instead of "Push Button" according to IEEE315-1975.

ubruhin:
Perfect :) Looks good now :+1:

PR #4, opened 2018-11-01
Add pkg for SOT353

dbrgn:
Regular:

screenshot

Hand soldering variant:

screenshot

Datasheet: https://www.nxp.com/docs/en/package-information/SOT353.pdf


pkg/0e89154b-a448-4fea-a115-169fb56c0cb9/package.lp
```diff
@@ -0,0 +1,185 @@
...
+ (created 2018-11-01T19:29:35Z)
+ (deprecated false)
+ (category b77c2754-fc91-4e9e-949e-5568a4bee0fe)
+ (pad 34f43b73-db56-4649-8904-0d258eeb7bfe (name "1"))
+ (pad a2b521d2-b58b-45b4-944a-65bd6e6e158a (name "2"))
+ (pad fa0eed07-e0ca-4ed6-bb33-4c54a46bb925 (name "3"))
+ (pad dc18d314-08ec-45f8-bd81-8b83857d1402 (name "4"))
+ (pad a837e245-00a8-45f7-befe-8aaafc1001f8 (name "5"))
+ (footprint 071e06d0-7e9d-4667-8a85-ab3dad74a72a
+  (name "default")
```
ubruhin: This is actually the reflow footprint, right? Then you should name it "reflow" to make this clear.

pkg/0e89154b-a448-4fea-a115-169fb56c0cb9/package.lp
```diff
@@ -0,0 +1,185 @@
...
+  )
+  (polygon f1ab1098-3aeb-4172-b9e3-25f7b70370a8 (layer top_documentation)
+   (width 0.1524) (fill false) (grab_area true)
+   (vertex (position 0.63 1.0) (angle 0.0))
+   (vertex (position -0.63 1.0) (angle 0.0))
+   (vertex (position -0.63 -1.0) (angle 0.0))
+   (vertex (position 0.63 -1.0) (angle 0.0))
+   (vertex (position 0.63 1.0) (angle 0.0))
+  )
+  (polygon f1ab1098-3aeb-4172-b9e3-25f7b70370a8 (layer top_placement)
```
ubruhin: This polygon makes the silkscreen overlapping with pads, which is bad. You should change it to what IPC7351 suggests (the inner rectangle):
grafik

Same for the other footprint.

pkg/0e89154b-a448-4fea-a115-169fb56c0cb9/package.lp
```diff
@@ -0,0 +1,185 @@
...
+   (position -1.2 0.0) (rotation 0.0) (size 1.2 0.4) (drill 0.0)
+  )
+  (pad fa0eed07-e0ca-4ed6-bb33-4c54a46bb925 (side top) (shape rect)
+   (position -1.2 -0.75) (rotation 0.0) (size 1.2 0.6) (drill 0.0)
+  )
+  (pad dc18d314-08ec-45f8-bd81-8b83857d1402 (side top) (shape rect)
+   (position 1.2 -0.75) (rotation 0.0) (size 1.2 0.6) (drill 0.0)
+  )
+  (pad a837e245-00a8-45f7-befe-8aaafc1001f8 (side top) (shape rect)
+   (position 1.2 0.75) (rotation 0.0) (size 1.2 0.6) (drill 0.0)
```
ubruhin: Aren't these pads for hand soldering too large? I would say it should be enough to make them just a little bit larger than the reflow soldering pads.

dbrgn: Note that they're 1.2x0.6mm with the large variant (with about 0.8mm sticking out under the lead), that's still tiny! Do you still think they should be made smaller?

pkg/0e89154b-a448-4fea-a115-169fb56c0cb9/package.lp
dbrgn: Fixed.

pkg/0e89154b-a448-4fea-a115-169fb56c0cb9/package.lp
dbrgn: Yes. But then it would be consequential to also name the regular footprint variants for all SMD parts (even with just a single variant) "reflow" :slightly_smiling_face:

pkg/0e89154b-a448-4fea-a115-169fb56c0cb9/package.lp
ubruhin: I'm pretty sure that 1.0x0.6mm should be enough for hand soldering (that's already 67% larger than the reflow solder pads).

pkg/0e89154b-a448-4fea-a115-169fb56c0cb9/package.lp
ubruhin: It's still extremely near to the pads and thus will probably overlap with the solder mask. So it needs to be even smaller. I would suggest at least 0.15mm clearance between pad and soldermask.

And now you also decreased the size of the polygon on the documentation layer :( Documentation should show the *real* size of the package, no matter if it overlaps with pads because it's not printed on silkscreen.

pkg/0e89154b-a448-4fea-a115-169fb56c0cb9/package.lp
ubruhin: Sure, the reason why other footprints aren't called "reflow" is just because most of them were imported from Eagle, and the importer doesn't know if a footprint is for reflow soldering. So it just calls them "default". Also LibrePCB names them "default" when creating a new package because LibrePCB also doesn't know it better.

pkg/0e89154b-a448-4fea-a115-169fb56c0cb9/package.lp
dbrgn: Since we explicitly have footprint versions with larger pads for hand soldering, why not make them actually easy to solder by hand? :slightly_smiling_face:  People that want space-efficient footprints will use the reflow version anyways.

But OK, I resized the pads to 1mm length.

ubruhin: 1.0x0.6mm might sound very small, but I guess this should be *easily* solderable by hand. And in my experience, space is nearly always a problem, so compact footprints are always a good thing :smiley: If there is still need for larger pads, we can also add a "hand soldering XL" footprint :grin:

pkg/0e89154b-a448-4fea-a115-169fb56c0cb9/package.lp
dbrgn: > Documentation should show the real size of the package, no matter if it overlaps with pads because it's not printed on silkscreen.

Can we have these things documented in guidelines? You seem to have a quite clear picture of how these symbols and footprints should look like, so please write them down. Otherwise I'll do all the work three times :slightly_smiling_face:

ubruhin: > Can we have these things documented in guidelines?

Sure, I just didn't find the time yet to do so...

In the mean time you can just take a look at the already existing library elements and do it the same way. For some of your new elements there are very similar elements already existing:
- SOT353 -> See SOT23-3
- ATtiny85 -> See STM32F103C
- Naming conventions (e.g. Title Case for element names), see *any* existing element

In addition, for packages please always take a look at IPC7351. I also don't know what is written there exactly, so I also take a look at it every time I review a package ;)

dbrgn: Done. I also changed the inner rectangle style to "filled", since that seems to be the IPC recommendation for SOT343.

dbrgn:
> Did you see #4 (comment)? ;)

No. You should unresolve conversations if you add a comment :wink: 

I forgot the "fixup" comment on the last commit, but you can squash all commits anyways.

dbrgn:
:tada: Thanks!

ubruhin:
Thank you :)

> You should unresolve conversations if you add a comment

Oh, I didn't realize that it was already resolved. It's a mess that GitHub automatically resolves discussions when pushing new commits :(

dbrgn:
It doesn't, it hides the discussion, but marks it as "outdated". I think I manually resolved that conversation before you added the comment.

PR #10, opened 2018-11-13
Add SOT23-5 and SOT23-6 packages

HBadertscher:
Hi, here are my first LibrePCB packages, I'd be happy to get some feedback

SOT23-5: 
sot23-5

SOT23-6:
sot23-6

Created according to Microchip's specifications for the [SOT23-5](http://ww1.microchip.com/downloads/en/DeviceDoc/05L_SOT-23_OT_C04-00091e.pdf) and [SOT23-6](http://ww1.microchip.com/downloads/en/DeviceDoc/6L_SOT-23_OT_OTY_C04-028C_OT.pdf).


dbrgn:
Here's a naming example from the recently updated SOIC parts:

screenshot

The keywords should contain all alternative / commonly used names btw.

HBadertscher:
Thanks for the feedback @dbrgn, I fixed the following things:

- Made the silkscreen outlines smaller, it should be exactly 0.15mm now
- Fixed the pin-1 dot as suggested. Actually it is not really required for the SOT23-5, but I left it in as it looks better imho and it doesn't hurt either. But I'm open to remove it completely.
- Centered the name and value. It seems be a more sensible default. Though you'll probably end up shoving those around manually for each PCB anyways.
- Added the leads on the TopDocumentation layer. I used the same sizes as in the SOT23-3 package and it looks reasonable to me.
- Changed names to IPC naming conventions
- Added other names to the keywords, names taken from [Wikipedia](https://en.wikipedia.org/wiki/Small-outline_transistor#SOT23-5,_SOT353)
- Bugfix: slightly misplaced pad in SOT23-5
- Bugfix: wrong pitch of the pads

About the IPC names: I named them `SOT95P280X145-x` as JEDEC only specifies a maximal height of 1.45mm. 

From what I have found, there are no recommended footprints available from JEDEC. I used the footprint recommended by Microchip and verified that the package outlines of Microchip conform to the JEDEC package outlines.

Updated pads:

sot23-5
sot23-6

ubruhin:
Thank you, looks good! :smiley:

PR #12, opened 2018-12-02
Add pkg for SOD123/SOD323/SOD523

dbrgn:
The IPC names are a bit less known than the JEDEC versions, but let's stick to them for consistency. The JEDEC name is mentioned in the description and tags.

Based on the following three datasheets from Diodes Incorporated:

- https://www.diodes.com/assets/Package-Files/SOD123.pdf
- https://www.diodes.com/assets/Package-Files/SOD323.pdf
- https://www.diodes.com/assets/Package-Files/SOD523.pdf

SOD3717X135 (SOD123) regular and hand soldering:

regular

hand

SOD2514X110 (SOD323) regular and hand soldering:

regular

hand

SOD1709X65 (SOD523) regular and hand soldering:

regular

hand


dbrgn:
Ping @ubruhin, this should be fairly quick to review :slightly_smiling_face:

ubruhin:
Sorry for the late review. To be honest, I'm not happy with the footprint variants (e.g. "default" or "reflow soldering" and "hand soldering") and thus I'm preparing package conventions which we can discuss then. As it's not only related to these diodes but for most of our packages, this PR can still be merged now - we would have to adjust most of our packages anyway after deciding for package conventions which are different than our current packages ;)

So, feel free to either merge this now, or to wait until the conventions are defined :)

Ah, I just noticed that in the commit message you write "SOT523", but in the package it's called "SOD523".

dbrgn:
> Ah, I just noticed that in the commit message you write "SOT523", but in the package it's called "SOD523".

Good catch :slightly_smiling_face: 

Yeah, we can still change the name of the variants later on.

PR #14, opened 2018-12-13
Add generic power connection symbol

pinoaffe:


ubruhin:
Thanks :)

My feedback:
- The category "Schematic Symbols" and its subcategories are intended for schematic-only symbols (like VCC or GND). But this is a symbol for "real" devices, so I would move it into the "Integrated Circuits" category.
- I would decrease the width to 10.16mm, this should be sufficient even for long pin names.
- Text elements should be left aligned (see [conventions](https://docs.librepcb.org/#libraryconventions-symbols-textelements)).

ubruhin:
I think now it's fine :+1: Or what do you think @dbrgn?

grafik

dbrgn:
My intuition would say that the pins should be oriented towards the right side for one-sided symbols. But I don't have a strong opinion on that. (The connectors are oriented towards the right though.)

Keywords are missing too. And should the category really be "integrated circuits"? Isn't there a power category? (Or is this not a power source? Would this be used for multi-symbol components?)

pinoaffe:
@dbrgn I wouldn't really know what keywords to add.
This is not a power source, it is indeed for use in multi-symbol components, as seen in https://github.com/LibrePCB-Libraries/LibrePCB_ICs.lplib/pull/5
I think it is in the correct category.

dbrgn:
> This is not a power source, it is indeed for use in multi-symbol components

Ah, then you're right of course! Maybe add a quick note in the description? Like "multi-symbol ICs" instead of "ICs"?

And re keywords, @ubruhin, the search will always consider the description and title too, right? Then we probably need no keywords.

ubruhin:
> And re keywords, @ubruhin, the search will always consider the description and title too, right? Then we probably need no keywords.

Correct.

So, let's merge this :)

Thanks @pinoaffe :+1:

PR #15, opened 2018-12-14
Add SMD chip resistors

dbrgn:
- IPC naming for chip resistors
- Height in the name is the max height (I pulled the most common resistor heights from Digikey)
- The smaller variants don't have silkscreen anymore
- Silkscreen lines are offset slightly, so that they should still be visible when the part has been mounted

3216 resistor:

img

0603 resistor:

img

5025 resistor:

img

Generated with https://github.com/LibrePCB/librepcb-parts-generator/pull/29.


ubruhin:
- Some of these packages already exist in this library -> please reuse UUIDs to overwrite them.
- Pad 1 is at the right, but should be at the left.
- Footprints "reflow" and "hand soldering" are exactly the same -> would be good to replace them by IPC density levels A/B/C anyway :)

dbrgn:
> Some of these packages already exist in this library -> please reuse UUIDs to overwrite them.

I'll try to do that, although it requires manually patching the UUID cache file.

> Pad 1 is at the right, but should be at the left.

Easy fix, thanks for spotting.

> Footprints "reflow" and "hand soldering" are exactly the same -> would be good to replace them by IPC density levels A/B/C anyway :)

Oops, I missed that one.

Regarding density levels, I found heel/toe/side guidelines in IPC and started implementing them. However, I don't think we can implement density level C in generic parts. The level C has very little tolerances, so it needs to match the part perfectly. (Remember, IPC specifies protrusion, not size.)

The thing is that we don't know the size of the coated leads of the chip resistors. The sizes are standardized, but not the length of the leads. So if we can't know the length of the lead, we can't really do well-matched footprints.

What about implementing footprint variants "Density Level A (Hand Soldering)" and "Density Level B (Reflow Soldering)" and leaving Density Level C to manufacturer specific libraries?

dbrgn:
Courtyard was added (according to IPC):

2012 footprint (level B):

img

2012 footprint (level A):

img

ubruhin:
> The thing is that we don't know the size of the coated leads of the chip resistors. The sizes are standardized, but not the length of the leads. So if we can't know the length of the lead, we can't really do well-matched footprints.

But then it might make sense to specify the lead length in our resistor packages (choosing the most commonly used lead length). Then we can add a density level C footprint, and these packages can be used for real devices if the lead length matches. In future we could also add more packages with different lead lengths.

> What about implementing footprint variants "Density Level A (Hand Soldering)" and "Density Level B (Reflow Soldering)" and leaving Density Level C to manufacturer specific libraries?

We can't move only one footprint to a manufacturer specific library, we would have to create a completely new package with density levels A and B specified again.

I noticed that you added packages with heights in 0.05mm interval. I think this is a far too small interval, it would lead to a huge amount of resistor packages. I would rather use an interval of at least 0.2mm. Then probably we only need one height per size to cover 90% of the available real resistors ;) 

Now I just found this sentence in the [German Wikipedia](https://de.wikipedia.org/wiki/Chip-Bauform):

> Die Höhe der Bauteile ist nicht standardisiert, bei Widerständen ist sie jedoch meist gleich, während sie bei Kondensatoren je nach Hersteller, Kapazität und Spannungsfestigkeit stark schwanken kann. 

So maybe it would make sense to even ignore the different heights completely (adding only one package per size, without specifying a height)? This would simplify things a lot, and for the 3D model a height error of e.g. 0.2mm doesn't matter IMHO...

> Some parts are not currently generated that were previously available as R-XXXX packages. I believe some of them are actually duplicates of others, but with metric name instead of imperial:

Oh f*ck, I didn't realize that (I imported them from a Eagle library) :sob: 

I would suggest to add the R-1218 and R-4527 as metrical variants 3246 resp. 11470 (or 11570?). But I'm not sure about the duplicate packages. I tend to remove them, although normally we shouldn't remove packages... We can just hope that nobody use them already ;)

dbrgn:
> We can't move only one footprint to a manufacturer specific library, we would have to create a completely new package with density levels A and B specified again.

Yes, but the different variants could then *exactly* match the manufacturer recommendations.

> I noticed that you added packages with heights in 0.05mm interval. I think this is a far too small interval, it would lead to a huge amount of resistor packages. I would rather use an interval of at least 0.2mm.

I used Digikey to find common heights. I only added them for variants which actually exist for thousands of different resistors, I didn't generate all combinations. (For example, the 1005 resistor only has 1 height variant.) Do you think the current amount of resistor packages is too much?

Generating 3D models later for different heights is trivial, but I would prefer it if the max height would actually match the part (for when you do very thin, space-constrained designs). And when considering mechanical constraints, then the height *is* important (much more so than the color).

> I would suggest to add the R-1218 and R-4527 as metrical variants

I can do that, won't hurt much.

> We can just hope that nobody uses them already ;)

Since the parts are copied to the project, it won't break anything, right? And they can be trivially replaced.

dbrgn:
Ok, branch updated.

I added 3246 (1218) footprints:

img

I did not add the 11470 (4527) footprint. These are packages that look different from other chip resistors, they have J-style leads and are quite rare. Here's an example:

https://www.newark.com/vishay/wsc45271k000fea/res-wirewound-1k-1-2w-4527/dp/65K4016

I don't think these should be generated with the same generator (I'd have to add quite a few conditionals because the leads are less wide than the package). I'd actually vote for removal of the package, I think the current one has pads that are too wide anyways.

dbrgn:
@ubruhin ping :slightly_smiling_face:

ubruhin:
OK, another idea: Probably most users (especially hobby users) will use "generic" resistor devices anyway (called e.g. "R-0805") because they don't care about which *exact* resistor will be assembled, and because it would take a lot of time to add all of the (thousands?) available (manufacturer-specific) resistors to our libraries. Thus maybe the best option is to make the corresponding packages for these devices also that generic that there is only one package per size, i.e. using only one specific lead length and one specific height (or even not specifying the height at all), without density level C footprint. So basically what you did now, but removing the packages with different heights.

In addition to these generic resistor packages and devices, specific packages and devices can be added step by step, with lead lengths and heights according the datasheets and so on. These are completely independent of the generic packages/devices, and thus out of scope of this PR.

So, a hobby user typically just uses the generic "R-0805" device with its generic "RESC2012 (0805)" package and assembles the random 0805 resistor he found in his resistor set. He doesn't care at all about lead length and package height, as long as he can hand-solder it and the 3D view looks nice.

On the other hand, a professional which might want to do a design with density level C and needs an accurate 3D export chooses the specific device of the exact resistor he evaluated, which has the corresponding package assigned, with exact lead length and package height, so it comes with density level C footprint and exact 3D model.

I think this would be a good compromise and suits both workflows very well. What do you think @dbrgn?

Btw:

> Since the parts are copied to the project, it won't break anything, right? And they can be trivially replaced.

Removing packages wouldn't break existing projects, but it would break existing devices. Meaning that you get an error when placing such a device on a board because the package can't be found.

dbrgn:
Yes, I think that sounds sensible!

It would be good to have another opinion from @rnestler or @HBadertscher, if one of you has time :slightly_smiling_face:

dbrgn:
> The nightly build shows clearance warnings between pads and placement layer

That's a float rounding issue. The silkscreen ends at X 2.15 and the pad starts at X 2.3. The difference (2.3 - 2.15) is 0.14999999999, so your tool complains.

I'll file a bug report (maybe I can also fix it). 

> IMHO we should clearly distinguish between generic and specific packages

You're right, I changed it. The max height was primarily meant for the 3D model, where we need to choose *some* height. I left in the parameters in the source file, but won't render it.

I also added a note to the description that the packages are generic.

> You also need to remove the corresponding devices as they are no longer valid ;)

Done.

ubruhin:
Looks very good now :+1:

The only thing which would be nice, is to see the actual lead size in the documentation layer, so the pad protrusion gets visually visible (at least the outer dimension, as the lead length of the generic packages aren't specified). Actually on the master branch this is already the case, but your change removes that "feature" :grin:

grafik

Would it be possible that add this again? (the glue rect in the center might also be nice, but let's ignore that for the moment)

I noticed that even the density level A pads are much smaller that the pads imported from Eagle, I wonder if they are not too small now for hand soldering :thinking:

dbrgn:
Updated doc layer.

img

ubruhin:
Only two things now (sorry :wink:):

The smallest resistors should have thinner lines on the documentation layer:
grafik

And the RESC2012 has very small leads, are you sure this is really intended? The leads are even narrower than those of RESC1608, which doesn't make sense to me. I checked a few datasheets, and they all have wider leads.

dbrgn:
I mostly used this datasheet:

img

2012 has a gap (B) of 1.4mm, 1608 has a gap of 0.8mm. That results in the mentioned difference.

So yes, it's intended. We could reduce the gap a bit (to be able to deal with a wider range of sizes) even though it feels a bit arbitrary.

I'll adjust the doc layer line thickness.

dbrgn:
Branch updated with latest changes from https://github.com/LibrePCB/librepcb-parts-generator/pull/29.

PR #16, opened 2018-12-18
Add DFN packages

HBadertscher:
So, I think the DFN packages are getting ready to merge. They are generated with LibrePCB/librepcb-parts-generator#22 - see that pull request for more information and discussions.

- Naming according to IPC wherever possible, plus certain additions which very similar to IPC. An example is `DFN50P300X300X95-10P30T220X160`:
  - `50P` = pitch 0.50mm (doesn't exist for DFN according to IPC, but would look like this for SOIC and QFN)
  - `300X300` = 3.00mm x 3.00mm package size
  - `X95` = nominal height=0.95mm
  - `-10` = 10 pins
  - `P30` = pad length 0.30mm (not IPC-suffix! If you know any better or official IPC suffix for this, please let me know!)
  - `T220X160` = thermal pad 2.20mm x 1.60mm. If the thermal pad is a square, then only one measure is given, e.g. `T220`.
- The silk-screen outlines have a line thickness of 0.254mm and exactly follow the border of the part. A dot of diameter 0.254mm marks pin 1. There are two different silk-screen variants: some have only two lines to indicate the top and bottom of the part, and for some the silk-screen also extends towards the pads, see print-screens below.
- Some pads and exposed pads are smaller than the nominal leads of the package. This is to make sure we always have a clearance of >= 0.20mm. The leads on the documentation layer show the correct nominal size. 
- There is always a "reflow" and a "hand-soldering" variant. The hand-soldering variant has 0.3mm longer pads. We'll have to see in practice, whether this value is perfect or if it should be tuned a bit.

- Some selected print-screens of the parts:
  - `DFN50P150X150X75-4T70X10` (reflow)
    dfn50p150x150x75-4t70x10

  - `DFN50P150X150X75-4T70X10` (hand-soldering)
    dfn50p150x150x75-4t70x10-handsolder

  - `DFN50P200X200X95-8T120X60`
    dfn50p200x200x95-8t120x60


dbrgn:
Very nice work!

> There are two different silk-screen variants: some have only two lines to indicate the top and bottom of the part, and for some the silk-screen also extends towards the pads, see print-screens below.

When do you use the first or the second variant? Depending on size / available space?

> A dot of diameter 0.254mm marks pin 1.

I wonder if the dot should be a bit larger, to make it easier to see. For example 0.512mm.

HBadertscher:
> When do you use the first or the second variant? Depending on size / available space?

Exactly. In the examples above, for the first package the silk-screen extends (a little) towards the pad, until it is 0.15mm from the pad. In the last case, this is not possible so I only draw a line (and move the line a bit upwards / downwards to keep the clearance of 0.15mm).

> I wonder if the dot should be a bit larger, to make it easier to see. For example 0.512mm.

For small packages, I think the smaller one looks better: compare the images above to that one: 

small_big_circle

For larger parts, why not:

big-big

How about some rule like for all parts with either W or L < 3.0mm we have the small dot and for all parts where both W and L >= 3.0mm we have the larger dot? (The second image in this comment is exactly 3.0mm x 3.0mm, i.e. the smallest part that would get a large dot)

dbrgn:
> How about some rule like for all parts with either W or L < 3.0mm we have the small dot and for all parts where both W and L >= 3.0mm we have the larger dot?

Sounds good to me.

Note: In the example above the point would collide with the name label. Note that the anchor point of the name aligns with the baseline, not with the lowest part of the text. Maybe you can avoid the collision by moving the dot down & left a bit.

ubruhin:
Awesome! :smiley: 

Is it intended that the height in the description is not given as nominal value too, like width and length? IMHO it would be less confusing if it's equal to the height given in the package's name.

In addition, I think it would make sense to also add the pad length to the description, at least for those packages which are available with different pad lengths. Otherwise it's not obvious what the e.g. "P30" in the package's name means.

Do the packages with an exposed pad with only 0.1mm width really exist (e.g. `DFN50P150X150X75-4T70X10`)? Somehow such a thin exposed pad looks strange ;)

dbrgn:
> Is it intended that the height in the description is not given as nominal value too, like width and length? IMHO it would be less confusing if it's equal to the height given in the package's name.

This is something I'm unsure about. From a mechanical design perspective, the max height is what matters. On the other hand, the nominal height is usually what's written out in the datasheet.

I used max height for the chip resistors (since there is no standardized nominal height), but I have the feeling that if there is a standard, then nominal height should be used in the package name and description. It's probably the designers job to account for tolerances.

ubruhin:
> I have the feeling that if there is a standard, then nominal height should be used in the package name and description.

:+1:

> It's probably the designers job to account for tolerances.

Yes I think so. In the end, LibrePCB is not a mechanical CAD, thus IMHO it's fine to ignore tolerances (if there is no standard defining them), and to allow small deviations in 3D models. Otherwise it's probably not possible to create libraries with generic packages without exploding the amount of packages (e.g. if we create a package for different heights in 0.05mm interval, it would lead to a huge amount of packages). For most designs, a 3D models accuracy of a few 0.1mm is totally sufficient. If a designer really needs high accuracy 3D models, he could still create his own library, with the *exact* 3D models for the particular parts he needs.

HBadertscher:
> Is it intended that the height in the description is not given as nominal value too, like width and length? 

You're right, I agree that this should be the nominal height, like in the name. I must have missed that when I changed the name from max to nominal. Fixed.

>  it would make sense to also add the pad length to the description, at least for those packages which are available with different pad lengths

Good idea, fixed.

> Do the packages with an exposed pad with only 0.1mm width really exist?

No idea, I've never seen them in the real world, but the JEDEC standard defines such exposed pads, so I just blindly created them ;-)

Finally, I now create small circles on the silkscreen for all packages where either length is < 3.00mm and larger circles if both lengths are >= 3.00mm. After tuning the circle location I think it looks reasonable. Below are examples with small and large circles.

Example with large circle
Example with small circle

ubruhin:
Looks good for me too, except that the generated files aren't canonical, i.e. LibrePCB modifies the files when saving those packages from the library editor. I would be fine with just running `workspace-library-updater` (located in the output directory after compiling librepcb) once over the packages, the generator script could be adjusted some day later to generate canonical files directly.

@HBadertscher If you're not familiar with `workspace-library-updater`, just tell me, I could run it quickly and push the resulting commit to your branch if you like :)

HBadertscher:
> the generated files aren't canonical, i.e. LibrePCB modifies the files when saving those packages from the library editor.

(Hopefully) fixed with a series of commits in the generator script. I manually tested a couple of packages and tried to run the `workspace-library-updater`, and to me it seems that the libraries should be canonical now. Could you verify that please?

ubruhin:
Perfect, looks good now! :+1:

Thank you very much for spending so much time to create these awesome packages! :smiley: :tada:

PR #17, opened 2019-01-05
SOIC packages: Add IPC density levels, add courtyard

dbrgn:
Use IPC density levels for SOIC footprints and add courtyard.

If the silkscreen clearance (0.15mm) would be violated, the heel is shrunk until the clearance is kept.

The documentation layer was resized to match the actual flat lead part. (This is what the density level offsets are based on.)

Generator PR: https://github.com/LibrePCB/librepcb-parts-generator/pull/30


dbrgn:
Looks like I'll have to re-add the projection of the leads according to the new guidelines [here](https://github.com/LibrePCB/librepcb-doc/pull/33/files) :slightly_smiling_face:

dbrgn:
Done.

ubruhin:
Nice! :+1:

In SOIC127P600X175-6 (and probably others, I didn't check all) the lead size in the documentation layer goes outside the copper pad, that's probably not intended?

grafik

dbrgn:
> that's probably not intended?

It is, the heel size is reduced to keep the 0.15mm clearance between silkscreen and copper.

By the way, both silkscreen and doc layer are *on* the boundary of the body. If we put the silkscreen entirely outside the body, then we'll have an even bigger problem with the clearances.

We could reduce the silkscreen to only top and bottom lines, although IPC has full rectangles in their land pattern document.

Furthermore, the documentation layer is also *on* the boundary, which is not fully according to the guidelines. But it would look a bit crappy if the silkscreen is on the boundary and the doc layer is inside:

img

In my opinion putting silkscreen and doc layer both on the boundary is the best tradeoff.

dbrgn:
I think this is now ready to review.

This is how a package looks like without documentation layer shown:

img

dbrgn:
I think this is now ready to review.

This is how a package looks like without documentation / courtyard layers shown:

img

PR #18, opened 2019-01-22
Add Dual MOSFET components

dbrgn:
- N/N
- P/P
- N/P


ubruhin:
For the P/N variant I'd suggest to rename the signals to make it clear which signals belong to which mosfet (e.g. "x1" -> "xN" and "x2" -> "xP").

In addition, all symbol variants should be adjusted a bit:
- All symbol items should have a suffix defined to make symbol names unique in schematics. I'd propose "A" + "B" for the N/N and P/P mosfets, and "N" + "P" for the N/P mosfet.
- One of the symbol should have a different position, otherwise both symbols overlap in the preview. Just set Y of the second symbol to -12.7 or so.

dbrgn:
> For the P/N variant I'd suggest to rename the signals to make it clear which signals belong to which mosfet (e.g. "x1" -> "xN" and "x2" -> "xP").

Done.

> All symbol items should have a suffix defined to make symbol names unique in schematics. I'd propose "A" + "B" for the N/N and P/P mosfets, and "N" + "P" for the N/P mosfet.

Done.

> One of the symbol should have a different position, otherwise both symbols overlap in the preview. Just set Y of the second symbol to -12.7 or so.

Done.

PR #20, opened 2019-01-25
Fix clearance exp-silk in certain DFN packages

HBadertscher:
Includes the changes from https://github.com/LibrePCB/librepcb-parts-generator/pull/31

Moves the silkscreen line a bit up/down as not to violate the clearance between the exposed pad and the silkscreen lines.


ubruhin:
Thanks @HBadertscher!

Btw, would be good to increment the version number when making changes to library elements. But let's ignore it this time :)

dbrgn:
I think version bumping should not be done in PRs, otherwise you'll get conflicts all the time.

dbrgn:
Ah, that's true :slightly_smiling_face:

PR #22, opened 2019-01-29
Add battery sym / cmp

dbrgn:
Single cell battery:

img

Multi-cell battery:

img

I categorized the items them both under "Schematic Symbols > Supply Symbols" and "Electromechanical > Battery Holders", since it could be both a battery holder or another type of connector.


ubruhin:
I would say all pins could be shortened by 2.54mm to make the symbols more compact. You can even remove the 2 horizontal lines by using the line of the pins instead ;) But when shortening the pins, the grab areas should be shortened too to avoid overlapping with the pins.

I'm not sure about the zero rotation of the symbols. Wouldn't it be more intuitive if they were vertically oriented, with plus at top (like capacitors)?

The components should have the default value "{{PARTNUMBER or DEVICE}}". What is the generic battery actually intended for?

> I categorized the items them both under "Schematic Symbols > Supply Symbols" and "Electromechanical > Battery Holders", since it could be both a battery holder or another type of connector.

The "Schematic Symbols" category (and subcategories) is intended for schematic-only symbols, i.e. not representing a real device, for example schematic frames or the "VCC" symbol. Thus the battery symbol is not correct here and should be removed from this category.

dbrgn:
> I would say all pins could be shortened by 2.54mm to make the symbols more compact. You can even remove the 2 horizontal lines by using the line of the pins instead ;) But when shortening the pins, the grab areas should be shortened too to avoid overlapping with the pins.

Done.

> I'm not sure about the zero rotation of the symbols. Wouldn't it be more intuitive if they were vertically oriented, with plus at top (like capacitors)?

Changed.

> The components should have the default value "{{PARTNUMBER or DEVICE}}".

Done.

> What is the generic battery actually intended for?

It can be used if you support single- and multi-cell batteries, or if you don't know how many cells there are.

But maybe that should be a generic device instead, that maps to the single-cell component?

> The "Schematic Symbols" category (and subcategories) is intended for schematic-only symbols, i.e. not representing a real device, for example schematic frames or the "VCC" symbol. Thus the battery symbol is not correct here and should be removed from this category.

Should it be added to another category besides "battery holders" (which isn't correct for LiPo connectors, for example)?

ubruhin:
Symbols:
- Are still in category "Supply Symbols" -> remove this category
- Name/value labels should be both on the same side. I would put them to the right for consistency with capacitors.
- The "+" and "-" signs overlap with netlines if netlines are connected from the right side to the pins. I would move the signs slightly nearer to the center of the symbol, and maybe make them slightly smaller to avoid such an overlap.
- Didn't we discuss offline that a "multi-cell" battery doesn't make much sense, and thus it should be renamed to "double-cell" instead?

> But maybe that should be a generic device instead, that maps to the single-cell component?

Yeah, I would say a generic battery component doesn't make much sense.

dbrgn:
> Are still in category "Supply Symbols" -> remove this category

Done. What about the question in my comment above? Should they be added to another category besides "battery holders" (which isn't correct for LiPo connectors, for example)?

> Name/value labels should be both on the same side. I would put them to the right for consistency with capacitors.

Done

> The "+" and "-" signs overlap with netlines if netlines are connected from the right side to the pins. I would move the signs slightly nearer to the center of the symbol, and maybe make them slightly smaller to avoid such an overlap.

Done

> Didn't we discuss offline that a "multi-cell" battery doesn't make much sense, and thus it should be renamed to "double-cell" instead?

I can't remember, that would be a few months ago. Are you talking about the symbol or the component? Why does it not make sense? (If we have a double-cell battery component, then why not 3-cell and 4-cell components too?)

dbrgn:
>> But maybe that should be a generic device instead, that maps to the single-cell component?
>
> Yeah, I would say a generic battery component doesn't make much sense.

I removed the component. But I can't add a generic device without choosing a package. I guess this should be left to concrete battery connectors. Alternatively we could add a generic battery connector device with a 2-pin "soldered wire connector" package.

dbrgn:
Thanks for the review @ubruhin!

PR #27, opened 2019-04-18
Switches, Relays

dbrgn:
Coil symbol:

coil

Switch SPDT 2-Position EU:

eu

Switch SPDT 2-Position US:

us

Relay DPDT Non-Latching with two symbol variants:

cmp

eu

us


dbrgn:
If this is good so far I can also add other relay variants.

dbrgn:
I added a 2-coil variant and removed references to latching functionality. Whether or not a single-coil relay is latching should be defined in the device, not in the component (since the signals are the same).

ubruhin:
Sorry for the delay, here my review :)

- Coil symbol: I would rename it to "Relay Coil / Solenoid" since "Coil" is too generic (many kinds of coils are represented with the inductor symbol or something similar).
- Switch symbols: Are the pin names A/B/C taken from some standard or convention? If not, I would prefer something more expressive like NO/NC/COM.
- Components:
  - Signal names: See note on pin names (NO/NC/COM).
  - The Rst+/- signals should be marked as "required" too, don't they?
  - Symbols do not have a suffix set. I would use the same suffixes as in the OpAmp components.
  - Symbol order is different to the OpAmps - for consistency it might make sense to move the coil to the end of the list, although of course coil/switches is not directly comparable to OpAmp supply/channels ;)
  - You marked all symbols as "required", but actually only the coil symbol is required. It's legal to have only some of the switch symbols placed in the schematic.
  - The order of symbols is different to the order of signals, i.e. the signals A1B1/C1 are mapped to the second switch, and A2/B2/C2 are mapped to the first switch:
grafik
  - The pin functions are clear from the symbols, so I would set "Designator in Schematics" to "None (no text)". They look ugly anyway in schematics ;)

dbrgn:
@ubruhin thanks for the review. All change requests should be implemented.

> Coil symbol: I would rename it to "Relay Coil / Solenoid" since "Coil" is too generic (many kinds of coils are represented with the inductor symbol or something similar).

:heavy_check_mark: Done

> Switch symbols: Are the pin names A/B/C taken from some standard or convention? If not, I would prefer something more expressive like NO/NC/COM.

:heavy_check_mark: Done

> Signal names: See note on pin names (NO/NC/COM).

:heavy_check_mark: Done

> The Rst+/- signals should be marked as "required" too, don't they?

Yeah, that probably makes sense.

> Symbols do not have a suffix set. I would use the same suffixes as in the OpAmp components.

Done. I used "R" for the reset coil.

> Feel free to decide on yourself whether to leave or change the order.

I would leave the coil at the top, since it's the most important part. I think it's more logical that way.

> You marked all symbols as "required", but actually only the coil symbol is required. It's legal to have only some of the switch symbols placed in the schematic.

Ah, I wasn't aware that this is possible! Very convenient. I removed the flag.

> The order of symbols is different to the order of signals, i.e. the signals A1/B1/C1 are mapped to the second switch, and A2/B2/C2 are mapped to the first switch:

:heavy_check_mark: Fixed

> The pin functions are clear from the symbols, so I would set "Designator in Schematics" to "None (no text)". They look ugly anyway in schematics ;)

:heavy_check_mark: Done

ubruhin:
Thanks! I rebased the commits to master.

PR #28, opened 2019-04-19
WIP: Use Travis-CI to check library

ubruhin:
Fails if invalid or non-canonical files were committed.

WIP because LibrePCB CLI 0.1.1 is not released yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,25 @@
+os: linux
+dist: xenial
+language: minimal
+
+addons:
+  apt:
+    packages:
+      - xvfb
+
+notifications:
+  email:
+    on_success: change
+    on_failure: change
+
+env:
+  CLI_VERSION: "nightly"
+
+install:
+  - wget -O ./librepcb-cli "https://download.librepcb.org/nightly_builds/master/librepcb-cli-$CLI_VERSION-linux-x86_64.AppImage"
+  - chmod a+x ./librepcb-cli
+  - xvfb-run -a ./librepcb-cli --version
+
+script:
+  - xvfb-run -a ./librepcb-cli open-library --all --save .
+  - git diff --exit-code  # fail if there are local changes


dbrgn:
Version 0.1.1 is released, and the tests pass!

Is this PR ready for review?

rnestler:
@ubruhin This would be awesome to have! A lot of time reviewers need to check recurring stuff that could be automated like this :slightly_smiling_face:

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+    packages:
+      - xvfb
+
+notifications:
+  email:
+    on_success: change
+    on_failure: change
+
+env:
+  CLI_VERSION: "nightly"
```
rnestler: Probably shouldn't use nightly, but a stable release, otherwise new lints on old devices will show up on unrelated changes.

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+
+install:
+  - wget -O ./librepcb-cli "https://download.librepcb.org/nightly_builds/master/librepcb-cli-$CLI_VERSION-linux-x86_64.AppImage"
+  - chmod a+x ./librepcb-cli
+  - xvfb-run -a ./librepcb-cli --version
```
rnestler: Does librepcb-cli need a xvfb to run?

ubruhin:
> @ubruhin This would be awesome to have! A lot of time reviewers need to check recurring stuff that could be automated like this slightly_smiling_face

It does not really check for library conventions since we can't suppress specific messages yet (which would be needed for checking library elements on CI). It's mostly useful for generated files to check whether they are canonical, i.e. LibrePCB would write them exactly identical.

> Should we maybe use CircleCI with a Docker image that already contains the librepcb-cli stuff? Or is this nod possible because of the xvfb requirement?

Actually I'd like either Azure Pipelines (because we already use that for LibrePCB itself) or the upcoming GitHub integrated CI (which is based on Azure AFAIK). I don't like to use too many different CI systems if not absolutely needed. 

And yes, I'd like to use a Docker image but unfortunately our official `librepcb-cli` image doesn't work on Azure because it has an entrypoint set (and maybe also because it's based on Alpine). Now I'm not sure what would be the best solution... Probably we would have to create an extra Docker image specifically for CI of libraries.

.travis.yml
dbrgn: Yes, since it runs a headless version of the GUI.

ubruhin:
Replaced by #50.

PR #30, opened 2019-06-15
Add more connector package categories

dbrgn:
Matching the component categories.


ubruhin:
- Names should be singular -> remove trailing "s"
- Author doesn't has to be "LibrePCB" so you could use your name if you like :)

dbrgn:
Thanks, fixed the singular names. Author name doesn't matter to me.

PR #33, opened 2019-07-07
Add TSSOP packages

dbrgn:
Generated with parts-generator, see https://github.com/LibrePCB/librepcb-parts-generator/pull/28

Based on JEDEC MO-153.


ubruhin:
Great, the long-awaited TSSOP packages! :smile: 

I didn't review them in detail yet, but so far I have some questions/comments:
- Regarding standard: Did you see [JEDEC MO-153](https://www.pcb-3d.com/wordpress/wp-content/uploads/SOP_JEDEC_MO-153F.pdf)?
- The name follows a different pattern than SOIC, *QFP and DFN packages. Is this intended, and if yes, what's the reason?
- From the description it's not clear what the last two numbers (after "L") in the name are actually meaning. Would be good to extend the description to explicitly mention these numbers.

The geometry itself is of course very nice since you could reuse the SOIC generator :)

dbrgn:
> Regarding standard: Did you see JEDEC MO-153?

Whoa, why is the JEDEC search totally useless? I searched for such a document for quite a while and didn't find it. Thanks! I updated all dimensions used according to JEDEC and added a few more packages.

> The name follows a different pattern than SOIC, *QFP and DFN packages. Is this intended, and if yes, what's the reason?

IPC7351B vs IPC7351C Draft. I used the latter. I also think it's the better approach.

We can still rename the other packages later on to become more consistent.

> From the description it's not clear what the last two numbers (after "L") in the name are actually meaning. Would be good to extend the description to explicitly mention these numbers.

It's lead length and width. I updated the generator.

PR #34, opened 2019-07-13
Add "RF Connectors" cmpcat

dbrgn:
For connectors in this list: https://en.wikipedia.org/wiki/List_of_RF_connector_types

Note: The matching packages are in the "Coaxial Connectors" category. Not sure if the mismatch is bad, but I thought that for devices it's logically better to have an RF connectors list than a coaxial connectors list.


ubruhin:
Hmm aren't basically all RF connectors coaxial? If yes, why not using the already existing "Coaxial Connectors" category?

dbrgn:
Ugh, I was on the wrong branch and did not notice that I [added that category myself a while ago](https://github.com/LibrePCB-Libraries/LibrePCB_Base.lplib/pull/30) :joy:

PR #36, opened 2019-07-31
Add sym and cmp for Switch DPDT

dbrgn:
Double-pole double-throw switches (e.g. slide switches).

Based on SPDT symbols, but with dashed line in between to symbolize the connected poles.


sym/58203f76-5cc0-4baf-aa1c-70026c53fa90/symbol.lp
```diff
@@ -0,0 +1,82 @@
+(librepcb_symbol 58203f76-5cc0-4baf-aa1c-70026c53fa90
+ (name "Switch DPDT 2-Position EU")
+ (description "Double-Pole Double-Throw switch, European symbol (IEC 60617).")
+ (keywords "switch,spdt,double-pole,double-throw")
```
ubruhin:
```suggestion
 (keywords "switch,dpdt,double-pole,double-throw")
```

dbrgn:
@ubruhin good otherwise?

ubruhin:
Honestly I'm not sure about the terminology. There are many different kinds of switches, and usually one wants to recognize it by the symbol in the schematic. That's why we have separate symbols/components for push buttons.

What device do you have in mind to be used with this component?

dbrgn:
> What device do you have in mind to be used with this component?

E.g. a slide switch:

image

Or one like this:

image

Both switches work with the same principle, only that in one case the switch is made of plastic, and in the other case it's a metal switching element.

https://www.littelfuse.com/technical-resources/technical-centers/commercial-vehicle-technical-center/poles-and-throws.aspx

A pushbutton is momentary and works differently, that's why IEEE gave it a different symbol.

Here are the IEEE symbols for switches:

img

There are a lot of different switch types and more specialized versions can still be added later.

ubruhin:
If I remember correctly, we discussed offline that the symbol for this kind of switches should contain the actuation type, i.e. hand-actuation in this case. Am I correct @dbrgn? :grin:

dbrgn:
Yes, but we still want the generic variant as well. We already have a matching generic SPDT version ("Switch SPDT 2-Position EU").

I can add more variants if you want, but that's in addition to the one that's already in this PR :slightly_smiling_face:

dbrgn:
@ubruhin I added a manually triggered variant with IEC symbol. No IEEE symbol so far, but that can always be added later on.

PR #38, opened 2019-08-12
Add rotary encoder

dbrgn:


ubruhin:
A slightly more expressive symbol would be nice, maybe like this one in KiCad?
grafik

dbrgn:
@ubruhin like this?

img

ubruhin:
Looks good, although I'd prefer to have the pins at the left side :grin:

And as discussed offline, it would be nice to have both symbol variants, one with the switch integrated into the encoder symbol and one with separate encoder and switch symbols (but we could still add that second symbol variant later).

dbrgn:
@ubruhin I moved the pins to the left and created a second symbol variant with integrated switch.

PR #41, opened 2019-10-05
added 10 bga footprints from lattice data sheets

ouabache:
added 3 pkgcats

Bga packages needed for Lattice iCE40 fpga family


dbrgn:
Cool, thanks for your PR!

The parts contain the note "Generated file---do not edit--". How did you generate them? Ideally we have a BGA generator that is used for all BGA parts, and that the community can maintain. We're still actively working on improving [librepcb-parts-generator](https://github.com/LibrePCB/librepcb-parts-generator/). maybe you can create a pull request there for your generator script?

Right now the packages don't fully conform to the [package conventions](https://docs.librepcb.org/#libraryconventions-packages) yet. A few things I noticed:

- There is no silkscreen (placement layer). Silkscreen should be added in a way that it's placed exactly outside the physical package, still visible after placing the part on the PCB.
- There should be a rectangle on the documentation layer that shows the boundaries of the physical package, it should be fully inside the physical package.
- The value is not aligned at the bottom center
- The value is not aligned at the top center
- The name should conform to the IPC7351C draft naming pattern:

screenshot

screenshot

Also, what's the difference between CSBGA/CABGA and BGA? Are they different, or just a special flavour of BGA?

ouabache:
The cs,ca and uc comes from the Lattice documentation. I enter the data
into a spreadsheet and use a perl script to create the packages.
It currently creates new uuids each time it runs so it really can't be
rerun once a library is released into the wild. Should I check it in? Where?

I'll fix the issues and submit another pull.

What do I do with the courtyard layer?

I will also add a pin A1 alignment dot.

John Eaton

ouabache:
Ok,

I fixed these issues and created a new pull request. Wasn't sure as to what
the difference was between "Collapsing vs Non-Collapsing" so I marked
everything Non-Collapsing. We may need to revisit this issue. I also need
to check the line width on the silkscreen to make sure it shows when the
part is placed.

John Eaton

PR #42, opened 2019-10-06
changed names to IPC7351

ouabache:
added silkscreen
added documentation rectangle
aligned name and value


dbrgn:
Hi @ouabache. You don't need to [close pull requests](https://github.com/LibrePCB-Libraries/LibrePCB_Base.lplib/pull/41) if you do a change, you can simply force-push (`git push --force`) to the remote branch if you do a change that overwrites the previous commit!

I'll answer here regarding the generator script:

> I enter the data into a spreadsheet and use a perl script to create the packages. Should I check it in? Where?

You could create a pull request for the script at https://github.com/LibrePCB/librepcb-parts-generator/. However:

> It currently creates new uuids each time it runs so it really can't be rerun once a library is released into the wild.

If stable UUIDs aren't possible right now, that would mean that we would always have to manually fix every one of those BGA footprints when we need to. That's a lot of effort, so stable UUIDs would definitely be desirable. (Usually it's not a big issue if we'd just need to remap a few pins, but in the case of BGA footprints we'd need to manually copy-paste dozens or even hundreds of pin UUIDs if we'd like to create a new version.)

Caching and a few other helpers ([UUID cache](https://github.com/LibrePCB/librepcb-parts-generator/blob/master/common.py#L15-L34), [formatting floating point numbers for LibrePCB](https://github.com/LibrePCB/librepcb-parts-generator/blob/master/common.py#L44-L55), [formatting dimensions according to IPC](https://github.com/LibrePCB/librepcb-parts-generator/blob/master/common.py#L58-L64) etc) are already implemented in the set of Python scripts that we use. What's the reason you didn't use those, is it just that you're not familiar with Python?

ouabache:
Hi Danilo,

Thanks for the tips on git. I use it for all my design work but this is the
first time that I have ever had to use branches. I only work in master.

Perl has been my main scripting language for 20 years and I use it when I
am in a hurry. I have used python but it takes a lot more work to get
results.

For now I will not check in any scripts until I solve the uuid issue. I
will generate the initial packages and then edit them for any changes.

Should I add a courtyard layer to the BGA packages?

John Eaton

rnestler:
@ouabache  What is the status on this?

> For now I will not check in any scripts until I solve the uuid issue.

Could you already solve it?

> Should I add a courtyard layer to the BGA packages?

Probably a good idea. What do @dbrgn @ubruhin think?

dbrgn:
Yes, courtyard would be good.

Honestly I'm still not sure whether it's a good idea if we start accepting complex generated parts to the standard library that aren't generated with the python scripts we use. If we decide to port the Perl code to Python in the future, we'd have to manually map all UUIDs to our cache file format, so that existing parts don't break.

It's a lot of duplicated effort (two generators) and might cause additional work in the future if we need to make adjustments to the generated parts.

ubruhin:
> Honestly I'm still not sure whether it's a good idea if we start accepting complex generated parts to the standard library that aren't generated with the python scripts we use. If we decide to port the Perl code to Python in the future, we'd have to manually map all UUIDs to our cache file format, so that existing parts don't break.

Yes, I totally agree!

ubruhin:
Closing this due to long time no progress and because such packages really should be generated with `librepcb-parts-generator`, otherwise they become unmaintainable for us. Sorry for keeping this PR open so long :-/

PR #43, opened 2019-10-11
Add more frames for ISO paper sizes

ouabache:
A0 and A1 paper sizes exceed default screen size


ubruhin:
Thanks! :smiley: 

FYI, I added a commit to bump the version of the modified component.

PR #44, opened 2019-10-26
Nixie tubes

dbrgn:
Add category, symbols and components for nixie indicator tubes (a version with decimal dot, a version without it).

image

I think it would be good to have those in the base library. The tubes themselves (packages and devices) probably belong into a "Nixie Tubes" library. (Creating manufacturer-specific libraries probably doesn't make much sense, those are all Russian tubes that are long out of production.)


ubruhin:
> I think it would be good to have those in the base library. The tubes themselves (packages and devices) probably belong into a "Nixie Tubes" library. (Creating manufacturer-specific libraries probably doesn't make much sense, those are all Russian tubes that are long out of production.)

I agree to have a separate library for tubes, but then I'd suggest to use it for all kinds of tubes (not only Nixie) and put the cmpcat/sym/cmp there too. I don't see why such legacy things should be downloaded by every LibrePCB user :wink:

dbrgn:
Alright :slightly_smiling_face:

dbrgn:
Replaced by https://github.com/LibrePCB-Libraries/Tubes.lplib/pull/1!

PR #49, opened 2019-11-22
Add molded polarized capacitors (CAPPM)

dbrgn:
I spent quite a while on this. Initially I used the IPC based generator for the pads, based on the size of the package and the physical gap. However, in contrast to chip resistors, the molded capacitors have an "inverted L" lead type that does not cover the full width of the package.

I ended up using the KEMET datasheet for their polarized tantalum capacitors to implement all these packages. That datasheet is very thorough and contains everything I needed (including pad sizes for all three density levels). I did not find a comparable datasheet for their competitor AVX.

https://content.kemet.com/datasheets/KEM_T2005_T491.pdf

Relevant pages: Page 4 (physical package) and page 15 (land pattern).

This means that the packages are not generated according to IPC formulas, but instead use the values specified by KEMET. But I think that is a good tradeoff, they probably sell billions of capacitors and should have good recommendations for the land patterns.

Since the lead size is part of the name (as per IPC751C), if a manufacturer has wildly different lead sizes for the same-sized part, we can always generate new packages. But at least for AVX, I compared 4 of the packages and the physical dimensions (lead length/width) were the same.

Here's CAPPM320X160X120L80X120:

cappm

The PR also contains some fixes for the RESC packages that resulted from changes in the generator.

Generated with librepcb-parts-generator: https://github.com/LibrePCB/librepcb-parts-generator/pull/64


ubruhin:
My thoughts:
- Why is there a space (in y direction) between documentation and outline polygons? In other packages we don't have that space...
- The documentation polygons have the grab area property set, maybe remove it like in #51?
- Somehow I'm not very happy with the outline because in x-direction it's not really an accurate guide where the package needs to be placed. What about something like this? 
grafik
(although nothing critical, i.e. I would also approve without changing it).

dbrgn:
> Why is there a space (in y direction) between documentation and outline polygons? In other packages we don't have that space...

True, that's probably a bug in the generator.

> The documentation polygons have the grab area property set, maybe remove it like in #51?

:+1:

> Adding the polarity indicator bar to the documentation layer would nice (to make assembly plans more expressive).

Good point!

> And maybe even add it to the outlines layer too?

The outline? Do you mean placement? That will always collide with the pad.

> Somehow I'm not very happy with the outline because in x-direction it's not really an accurate guide where the package needs to be placed.

The U-shape is the IPC recommendation:

img

Although it could be argued that since the pads extend a bit outside the outline, it's similar to this one:

img

I'd keep it for now, we can always adjust the placement layer if it doesn't look good on real PCBs.

ubruhin:
>>  And maybe even add it to the outlines layer too?

> The outline? Do you mean placement? That will always collide with the pad.

Yes, I mean placement :) Of course it would need to be drawn either at top/bottom or to the right of the pad to avoid collision. But OK, let's draw it only on the documentation layer.

> I'd keep it for now, we can always adjust the placement layer if it doesn't look good on real PCBs.

:+1:

dbrgn:
> Why is there a space (in y direction) between documentation and outline polygons? In other packages we don't have that space...

Fixed!

> The documentation polygons have the grab area property set, maybe remove it like in #51?

Good point, however that will also affect chip resistors. I'll do that in a separate PR.

> Adding the polarity indicator bar to the documentation layer would nice (to make assembly plans more expressive).

Added to doc layer:

pol

@ubruhin I think this should address all open issues.

ubruhin:
> I think this should address all open issues.

Thanks!

> Good point, however that will also affect chip resistors. I'll do that in a separate PR.

Since this branch already contains two commits affecting only the resistors, could you maybe first create the separate PR to update the resistors, so these two commits will disappear from this branch after rebasing? Would be a much cleaner history IMHO :)

And the newly added polarity bar also has the "grab area" set, will it be removed too after the update?

Btw. I just noticed that the KEMET/AVX case codes are not contained in the keywords, thus we won't find the proper packages when searching e.g. for "kemet x". Might be nice to add these terms to the keywords since those (at least the KEMET codes) are much better known than the EIA size codes AFAIK...

Otherwise the packages look good :+1:

dbrgn:
> Since this branch already contains two commits affecting only the resistors, could you maybe first create the separate PR to update the resistors, so these two commits will disappear from this branch after rebasing? Would be a much cleaner history IMHO :)

I moved the two commits to a separate PR. I don't have the motivation to do the grab area change in that PR as well though, I've already spent a lot of time on this PR and don't want to do even more tweaking of the commit history both in this repo and in the parts generator.

> And the newly added polarity bar also has the "grab area" set, will it be removed too after the update?

Since it's a filled area and not an outline, the grab area property was intentional. But it probably makes sense to remove it as well.

> Btw. I just noticed that the KEMET/AVX case codes are not contained in the keywords

I updated the parts generator to make keywords dynamic. This PR is updated with the change (but needs rebasing once #54 is merged).

ubruhin:
> I moved the two commits to a separate PR. I don't have the motivation to do the grab area change in that PR as well though, I've already spent a lot of time on this PR and don't want to do even more tweaking of the commit history both in this repo and in the parts generator.

Hmm now I'm confused. You wrote to disable the grab area of the documentation layer:

>> The documentation polygons have the grab area property set, maybe remove it like in #51?
>
> Good point, however that will also affect chip resistors. I'll do that in a separate PR.

I expected the following procedure:

1. Disable the grab area in the generator script
2. Open a PR to commit the updated resistors to the base library
3. Rebase this PR (molded caps) to master -> would have *automatically* removed the two commits affecting the resistor packages, since these changes are already on master.

But now I see you opened #54 with the two commits from this PR, but without the actual change to the documentation layer. So it would require a third PR to also update documentation layer later, or did I misunderstood something?

It wasn't my intention to generate more work for you - my idea was just to do the work in a different order - first update the resistors, then add the molded caps (same work, but results in a cleaner history).

dbrgn:
The reason was that the grab area change should theoretically result in a version bump. The parts generator PR already contains such a version bump in this commit:

    ec491ce - chip: Bump version due to changed grab area on courtyard (2 weeks ago Danilo Bargen)

So to actually do this fix cleanly, I'd merge the grab area change into that commit. However, I cannot just change the grab area properties and then squash that change into the original commit, because due to refactorings that will result in a lot of conflicts. Instead, I'll have to go back to that point in time, apply the change into the commit, and then resolve potential conflicts in every following commit. Finally, I'll have to undo all changes in LibrePCB_Base.lplib, step through every commit in the parts-generator branch and then re-create the commit with the generated files in LibrePCB_Base.lplib (because I want separate changesets for the commits in the parts generator).

To avoid this, there are two alternatives:

- Create a new commit that does the grab area changes, and bump the version again. That's not really a much cleaner, linear history though, since there are two version bumps in the same PR.
- Do the changes at the end of the history without another version bump, since the changes in the commit that actually bumped the version were never merged and will thus not create any problems. But that's much uglier, because now the parts generator commits don't correspond to LibrePCB_Base.lplib commits anymore.

Anyways, we've both wasted more time than necessary writing those big comments. I'll pick the first alternative and will update the other PR with the grab area change, followed by a rebase of this branch.

dbrgn:
Both PRs updated.

dbrgn:
> But that's much uglier, because now the parts generator commits don't correspond to LibrePCB_Base.lplib commits anymore.

After second thought, that was already the case before I added the new commit :sweat_smile: There was a version bump commit, and then further changes without an additional bump commit.

So now that we have the grab area commit which bumps the version, I guess it's better than before :wink:

dbrgn:
Rebased, I hope everything's correct now :sweat_smile:

PR #50, opened 2019-11-30
Use GitHub Actions to check library

ubruhin:
Replaces #28.


dbrgn:
Nice! Runs quite fast.

PR #52, opened 2019-12-05
Add sym/cmp for solder jumpers 2P/3P

ubruhin:
The symbols are similar to those in Eagle and KiCad libraries.

grafik


dbrgn:
Do you already have packages for these? :slightly_smiling_face: Could be useful in the Gfrörli PCB.

ubruhin:
> I assume the components don't need a default value, right?

Correct, I think there is no reasonable value we could set for a jumper...

> Do you already have packages for these? slightly_smiling_face Could be useful in the Gfrörli PCB.

Yes I do, but they need some improvements to make them ready for inclusion in official libraries. I'll try to do it soon, or at least open a WIP PR.

rnestler:
Is this ready to merge?

PR #53, opened 2019-12-05
Add sym/cmp for dual suppressor diode bidirectional

ubruhin:
For example for [SM712](https://www.littelfuse.com/~/media/electronics/datasheets/tvs_diode_arrays/littelfuse_tvs_diode_array_sm712_datasheet.pdf.pdf) which is basically *the* protection diode for RS485 lines.

grafik


ubruhin:
> Is "in Series" the correct, commonly used term? 

Actually not really, I think there is no term at all for this kind of diode array (and the term "array" doesn't really help since it is way too generic - the term "dual" already implies that it's an array consisting of two diodes).

> To me, that terminology seems a bit confusing since they're both parallel (with a common pin).

Electrically they are not parallel, they are in series.

But maybe we could just remove the term "in series" since probably we'll never add any other array configuration of this kind of diode. If we do some time, we could still adjust the names to distinguish between the different variants. Do you agree @dbrgn?

dbrgn:
Yep, I agree :slightly_smiling_face:

cmp/20f51f1d-ba6d-4319-961e-85645888661f/component.lp
```diff
@@ -0,0 +1,33 @@
+(librepcb_component 20f51f1d-ba6d-4319-961e-85645888661f
+ (name "Suppressor Diode Bidirectional Dual In Series")
```
rnestler: Why not use the name from the datasheet? "TVS Diode Arrays"?

ubruhin: Because it is way too generic. See https://github.com/LibrePCB-Libraries/LibrePCB_Base.lplib/pull/53#issuecomment-562758022.

ubruhin:
Removed "In Series" from the name.

PR #54, opened 2019-12-07
Chip resistor updates

dbrgn:
The first two commits of #49, as requested there.


pkg/1596e275-314e-4bde-a5b5-11ea19d9d6e8/package.lp
```diff
@@ -3,7 +3,7 @@
  (description "Generic J-lead resistor 11569 (imperial 4527).\n\nLength: 11.56mm\nWidth: 6.98mm\n\nGenerated with librepcb-parts-generator (generate_chip.py)")
  (keywords "11569,4527,r,resistor,j-lead,generic")
  (author "Danilo B.")
- (version "0.3")
+ (version "0.3.1")
```
rnestler: Didn't we say we only use major.minor for the version?

dbrgn: I'm not aware of that. My proposal is still to use only integers, and increment them by 1 on any change. Right now the decision whether to use major, minor or patch version is arbitrary, since semver doesn't make sense for parts.

ubruhin:
I'm really sorry, but I guess there is still something wrong with the changes... I noticed that the commit "Chip resistors: Disable grab area on doc layer" also changes the courtyard coordinates of the "Density Level A" footprint, but not of the "Density Level B" footprint. Thus I checked out the `chip-cap-unip-pkg` branch of the generator and re-generated the resistor packages. And indeed, this strange change in one of the footprints gets reverted, i.e. this branch does not correspond to the current state of the generator branch `chip-cap-unip-pkg`. Or is this intended?

dbrgn:
Hm, when I go back to the commit "chip: Disable grab area on doc layer" and re-generate, there are no changes on master... Maybe something went wrong when rebasing and fixing the conflicts.

Rebasing branches that also contain refactorings and keeping generated data in another repository in sync is a pain :slightly_frowning_face:

dbrgn:
@ubruhin I now applied the diff generated by the current HEAD of the generator branch, as suggested.

PR #58, opened 2020-01-26
Rework DIP packages

dbrgn:
This is a total rework of DIP package generator. Another one of those PRs where I thought "let me quickly do those changes", but then I spent a total of 10-15 hours on it :sweat_smile:  (Mostly research)

See https://github.com/LibrePCB/librepcb-parts-generator/pull/75 for more details. Regarding the data used for generating the updated packages, please take a look at [the docstring of the generator script](https://github.com/LibrePCB/librepcb-parts-generator/blob/10214e1f15a8ba278274135e3e96e9959cc59f3c/generate_dip.py#L1-L159). I spent quite some time collecting data about DIP packages.

Changes:

- IPC 7351-C conforming names
- Updated dimensions (based on JEDEC and real-world package sizes)
- New silkscreen conforming to IPC 7351-C recommendations
- Use lead span for pad x-offset, not package width
- Added courtyard (0.4 mm excess)

Four DIP packages have been deprecated: DIP-36, DIP-48, DIP-52 and DIP-64. They have practically no real-world use, and if they do the packages sometimes deviate significantly from the standard sizes, so generic packages shouldn't be used.

Version has been bumped to v0.2.0.


dbrgn:
Hmm, I'll fix the non-canonical file issues.

dbrgn:
The problem was the version, LibrePCB prefers `0.2.0` over `0.2`. Not sure if that's a good idea, but that's how it is :slightly_smiling_face: 

I also added courtyards to the packages. Ready for review.

ubruhin:
Nice! :+1:

> Use lead span for pad x-offset, not package width

Although this might sound reasonable, it's *very* strange to have DIP packages with a pad distance which is not a multiple of 2.54mm. In probably every EDA library DIP packages have a pad distance of 7.62mm or 15.24mm. And these packages also fit on breadboards with 2.54mm grid. 

> The problem was the version, LibrePCB prefers 0.2.0 over 0.2. Not sure if that's a good idea, but that's how it is slightly_smiling_face

That's to make the file format more canonical by disallowing different representations for the same version number ("0.2" and "0.2.0.0.0.0" are different, but represent the same version number) :wink:

dbrgn:
Finally updated :joy:

ubruhin:
Rebased and re-generated with the changes from https://github.com/LibrePCB/librepcb-parts-generator/pull/75#issuecomment-1703387080.

In addition, I quickly resolved all warnings on the deprecated DIP packages. The auto-generated package outline and courtyard polygons are not as accurate as those from the parts generator, but I think that's good enough for deprecated packages and better than just approving all warnings.

PR #62, opened 2020-04-01
Add sym/cmp for resistor array/network 4..8x

ouabache:
##### SUMMARY

added resistor pacs
##### CHECKLIST


dbrgn:
Hi, thanks! I took a quick (but not complete) look.

- For the symbols, as far as I'm aware there is not really an IPC standard so we currently stick to regular names (e.g. `Resistor Network 8x EU`) instead of IPC style names (`RES_NETWORK_8_EUR`).
- The resistor arrays should probably not be a single symbol, but you should probably create a component that is composed of 8 standard resistor symbols
- You added components in a nested subdirectory (`cmp/cmp/`), these cannot even be opened by LibrePCB
- A generic resistor array or network device probably doesn't make much sense, since in practice packages used by vendors differ. A resistor network or array is not a standard component like a resistor is. Therefore I'd remove the devices, they should be actual products that can be checked against a datasheet.
- You did not yet accept the CC0 Public Domain license for your contribution yet (see pull request description above)

ouabache:
On Wed, Apr 1, 2020 at 1:59 PM Danilo Bargen <notifications@github.com>
wrote:

> *@dbrgn* requested changes on this pull request.
>
> Hi, thanks! I took a quick (but not complete) look.
>
>    - For the symbols, as far as I'm aware there is not really an IPC
>    standard so we currently stick to regular names (e.g. Resistor Network
>    8x EU) instead of IPC style names (RES_NETWORK_8_EUR).
>
>
fixed

>
>    - The resistor arrays should probably not be a single symbol, but you
>    should probably create a component that is composed of 8 standard resistor
>    symbols
>
>
I did both. There is a symbol variant using 8 discreet parts. Should I make
that the default?

>
>    - You added components in a nested subdirectory (cmp/cmp/), these
>    cannot even be opened by LibrePCB
>
>
 My bad

>
>    - A generic resistor array or network device probably doesn't make
>    much sense, since in practice packages used by vendors differ. A resistor
>    network or array is not a standard component like a resistor is. Therefore
>    I'd remove the devices, they should be actual products that can be checked
>    against a datasheet.
>
> Deleted

>
>    - You did not yet accept the CC0 Public Domain license for your
>    contribution yet (see pull request description above)
>
>
Checked it.

dbrgn:
Great, thanks for the fixes! Now that the pull request is even smaller (devices and packages removed) that makes it even easier to review :slightly_smiling_face: 

> > The resistor arrays should probably not be a single symbol, but you
> > should probably create a component that is composed of 8 standard resistor
> > symbols
> 
> I did both. There is a symbol variant using 8 discreet parts. Should I make that the default?

Ah, nice!

@ubruhin what is your opinion on this? Should there be any "combined" resistor symbols?

For the network (where all outputs are connected) it probably makes sense to have a dedicated symbol. For the array, I'm not sure. A dedicated symbol may be easier to handle in the schematic editor, but it's not strictly needed and constitutes some degree of duplication.

@ouabache some more thoughts:

- Can you rename the pin 1 in the network symbols and components to "COM" and re-number the other pins? That would be more consistent with other symbols (e.g. "Switch SPDT 3-Position EU").
- Is the designator `RP` for resistor arrays/networks common? Do you have a source for that? (Couldn't find anything suitable in the IEEE spec...)
- The two components should have a good description as well as keywords (you can re-use the keywords of the symbols, I think those are fine)

ouabache:
I have never seen anything official but thats whats I have always used.
Everything else is fixed and checked in.

ubruhin:
Hi,

I have now also reviewed the symbols and components. My thoughts:

## Symbols

- The polygons all have the same UUID -> they should have different UUIDs.
- A symbol which consists of multiple single symbols looks confusing IMHO. In schematics it looks like all those resistors are separate symbols, but you can't drag them individually. Therefore I think there should be a box (filled with yellow background) around these symbols, then it's clear that they represent a single component.
- I would make the symbols more compact by using only 2.54mm pin distance instead of 5.08mm. The resistor symbols could be made a bit smaller to ensure enough clearance between each other.
- Would be nice to have junction points to explicitly mark the junctions (nice to have, no strong requirement).

In Eagle I found a library which also follows that style:

grafik grafik

## Components

- In the "Resistor Network 8x", the split symbol variants do not make sense to me. It would allow to put 8 independent(-looking) resistors to the schematics, but they aren't independent - if you connect pin 2 of one resistor, pin 2 of the 7 other resistors are implicitly connected to the same net, so they are not useful at all. Actually I'm not even 100% sure if the schematic editor can handle this situation properly, but that's a different topic...
- The "norm" attribute of symbol variants should be set to "IEC 60617" (EU) resp. "IEEE 315" (US) so LibrePCB is able to pick the variant which is configured in workspace resp. project settings.

Thanks!

ouabache:
Fixed.

Did not add the dots. They looked to big once I had shrunk the components
and I never did like dots on three way junctions

ouabache:
I am generating these symbols with a python script so I can easily create
an entire family with  different number of resistors. Let me know which
ones you would like to see.

I also tried out the split network on a board and schematic capture handles
it with no problem. It will give you an error if you try to connect a COM
pin to a different net than the first one

>

ubruhin:
Well done, I like the new look of the symbols! :+1:

Only two small issues are remaining:
- The symbol "Resistor Array 4x EU" has two overlapping outline polygons, therefore the yellow color looks darker than usual (see image below).
- The texts are not aligned to the outline -> according library conventions, they must be placed exactly on the edges of the outline polygon.

grafik

> I also tried out the split network on a board and schematic capture handles
it with no problem. It will give you an error if you try to connect a COM
pin to a different net than the first one

I thought again about this, and also tried it in the schematic editor. And I still think it is very weird to have multiple pins for the same component signal. Consider this situation:

grafik

What purpose do the 3 unconnected pins on the right side have? IMHO they are only confusing and do not provide any advantage. So I would *really* remove this symbol variant in the resistor network components. Or what do you think @dbrgn?

*If* we really need a "split" symbol variant, we should use resistor symbols with only one pin instead - then we don't have duplicate pins. But I'd still prefer to simply remove that symbol variant. We could still add it later if *really* needed.

ouabache:
Double polygon and text aligns are fixed and checked in.

In your example if you placed a network as individual resistors then you
would connect VCC to all of the pin 2s. You don't want to have to search
the schematic to find which supply a signal is pulled up to so we show an
explicit connection on each resistor. These connections have no effect on
the netlist but they should be included. If you tried to connect to any
other signal then you would get an error.

ubruhin:
Sorry for the long delay regarding this PR. I have now allowed myself to rebase and update this branch with some polishing, I hope that's fine :see_no_evil: 

The symbols now look like this:

image

However, I'm sorry but IMHO the split resistor networks are too confusing in a schematic. It's very strange and unusual to have the same physical pin represented by multiple symbol pins. Thus I removed the split symbol variants from the network components (not for the arrays). I assigned new component UUIDs to avoid breaking your existing designs.

If there's no more discussion on this, I'll merge it soon.

Thanks!

ouabache:
You can have a resistor network where the common connection is inside the
package or it can be made outside the package with a trace. The top two are
an example of a normal network with the common connection made on the PCB.

ubruhin:
I think I was not clear enough about my statement, sorry :see_no_evil: I was not talking about the resistor network *symbol*, but the split representation as a *symbol variant* in the component (see previous discussion in https://github.com/LibrePCB-Libraries/LibrePCB_Base.lplib/pull/62#issuecomment-612261531). So I didn't remove the symbols, just the *symbol variants* from the components.

PR #67, opened 2020-04-15
Add sym for two optical rotary encoders

ouabache:
##### SUMMARY

added two optical encoders

##### CHECKLIST

- [x ] I have read and followed the [library conventions](https://docs.librepcb.org/#libraryconventions).
- [ x] I'm the copyright owner of the added content (i.e. the changes are made by myself, not copied/imported from somewhere else).
- [ x] I agree to publish all my changes under the [CC0 Public Domain License](https://en.wikipedia.org/wiki/CC0), allowing everyone to use and modify the content without any restrictions.


ubruhin:
Generally, looks not bad to me :+1: But I still have some thoughts:

- Maybe remove the "CH_" prefix from pin names ("A", "B", "I" should be clear enough IMHO)?
- The symbol should represent an optical *rotary* encoder, right? Then I would say the term "Rotary" should also be contained in the name, e.g. "Encoder Rotary Optical".
- The symbol is unnecessary large and irregular. See the "Encoder Rotary With Switch" symbol how it should look like to be compact. Of course you can make it slightly wider (but not higher) because of the longer pin names.

ouabache:
Data sheets include the ch in the names so I copied that.

They are all actually linear encoders  but can work with a wheel if the
radius is large enough.

I will see if I can do something with the shape

ubruhin:
I applied the changes by myself now. Thanks for the contribution :tada:

PR #75, opened 2020-06-14
Adding generic 2-pos SPDT & concrete SS12D00G3/CYT1024 switch

sixtyfive:
- [x] I have read the library conventions
- [x] I'm the copyright owner of the added content
- [x] I agree to publish all my changes under the CC0 Public Domain License

This thing is [available online under various names](https://www.amazon.de/s?k=SS12D00G3&__mk_de_DE=%C3%85M%C3%85%C5%BD%C3%95%C3%91&ref=nb_sb_noss). It was not originally intended to be used as an SMD part but works fine that way. At some point it would be nice to add a THT package for it, but right now I don't have one.

Known-working.


dbrgn:
Blocked by #79.

PR #81, opened 2020-07-12
Add hall sensors as category

Luz:
##### SUMMARY
Add "Hall Sensors" as new subcategory to the category named "Sensors ..."

##### CHECKLIST

- [ ] I'm the copyright owner of the added content (i.e. the changes are made by myself, not copied/imported from somewhere else).


dbrgn:
@ubruhin should hall sensors be a sub-category of a new "Magnetic Sensors" that might also include things like reed switches or electronic compass ICs?

Luz:
I suggest to not make too many sub-sub-categories, since after a while you do not find the parts anymore. As long as we do not have hundreds of top-categories, I suggest to keep it simple.

ubruhin:
> I suggest to not make too many sub-sub-categories, since after a while you do not find the parts anymore. As long as we do not have hundreds of top-categories, I suggest to keep it simple.

Full ACK! But shouldn't we call the new category simply "Magnetic Sensors"? Even DigiKey has no "Hall Sensors" category. Actually hall sensors is not the sensor type, but the used technology. A hall sensor can be a current sensor, a compass, proximity sensor, ... IMHO categories should group components by function, not by technology.

PR #83, opened 2020-08-17
Add optical sensors category

dbrgn:
##### SUMMARY

Add a new sensor category "Optical Sensors".

None.

##### CHECKLIST


ubruhin:
IMHO the subcategories are too detailed, i.e. clutter up the category tree too much. At least as long as we don't have hundreds of components in these categories :wink: Keep in mind that thanks to generic components, the categories don't have to be very detailed. For example, we don't have separate categories like "Fixed Resistors" and "Variable Resistors", these are just separate components within the "Resistors" category. I'd say we should handle optical sensors in a similar way.

dbrgn:
@ubruhin updated!

dbrgn:
@ubruhin BTW, if you just leave a comment and not a review (even if you're requesting changes), I cannot re-request a review.

ubruhin (approved):

PR #86, opened 2020-12-26
Add TSOP-II packages from JEDEC MS-024

Tubbles:
##### SUMMARY

Add TSOP66P65 package. Note that since the std doc (JESD21-C I think) is not yet found, I cannot generate more packages in the TSOP range, and there is a slight risk that this package actually contains errors.

Note also that the TSOP package family is not mentioned in IPC7351C

##### CHECKLIST


ubruhin (approved):
Awesome, thanks! Looks good to me :smiley:

PR #89, opened 2021-04-28
Add motor component categories

dbrgn:
- Motors, Actuators
  - AC Motors
  - DC Motors
  - Stepper Motors

(Not sure if these should be on the root level, or below "Electromechanical". I put them on the root level for now.)


ubruhin (changes requested):
Sorry, now I have to request changes :grin: 

> Motors, Actuators
>
>    AC Motors
>    DC Motors
>    Stepper Motors

This looks much too detailed for me. For example for diodes we don't have "Schottky Diodes", "Suppressor Diodes", "Zener Diodes" etc., these are just the different components within a single "Diodes" category. I think we should handle different motor types the same way.

> (Not sure if these should be on the root level, or below "Electromechanical". I put them on the root level for now.)

I would put it under Electromechanical ;)

dbrgn:
@ubruhin updated!

ubruhin (approved):
Thanks!

PR #91, opened 2021-04-28
Add sym/cmp for reed switches

dbrgn:
##### SUMMARY

Normally open:

image

Normally closed:

image

##### DATASHEETS / REFERENCES

I did not find an official IEEE / IEC symbol in the documents I have, but googling for "reed switch symbol" returned mostly the same style of symbols.

##### CHECKLIST


dbrgn:
@ubruhin any chance for a review? It's a fairly simple PR.

ubruhin:
Generally the symbols look very large to me. Most 2-pin symbols have a width of 4 grid intervals, compared to 6 intervals of the reed switches:

image

I suspect the symbols would look more natural and consistent if they are only 4 intervals width as well, and less high. Would that be possible? :) Maybe for consistency reasons (since there doesn't seem to be a clear standard) it would even make sense to use the same symbol as the relay contacts, just adding the frame around it? IMHO the NC reed symbol is not really expressive (I'm not sure if I'd understand that symbol without knowing this PR) while the NC relay contact is very clear... But yeah, your variant seems to be used more often :(

In addition:
- The pin names should be hidden in the component since the polarity doesn't matter and the names overlap with the lines.
- I'm struggling a bit with the names, wouldn't e.g. "Reed-Switch NC" sound much more natural here? I know we have the guideline to put the generic terms to the front, but I'm not sure if this makes sense here - IMO reed switches are basically a completely own thing, not really related to other switches. Or what is your feeling about this @dbrgn?

dbrgn:
How about this? It looks a bit asymmetric, that's why I originally picked the symmetric version.

screenshot-20220420-001808

dbrgn:
> I'm struggling a bit with the names, wouldn't e.g. "Reed Switch NC" sound much more natural here?

Yeah, that probably makes sense.

dbrgn:
> IMHO the NC relay contact is even much more expressive than the reed switch (I'm not sure if I'd understand that symbol without knowing this PR).

It has 3 terminals though, not 2. The suggestion above reuses the same concept, maybe this is what you meant?

(The small symbol with the original symmetric approach would probably work as well.)

ubruhin (approved):
Sorry for the long delay! Although probably there's still some room for visual improvements of the symbols, at least they now have a pin distance of 4 grid intervals so we could improve the look some day later without possibly breaking any schematic :smiley: So feel free to merge this if it is ready.

PR #92, opened 2022-02-26
Add 5mm and 3mm generic LED packages & devices

dbrgn:
##### SUMMARY

This adds generated packages for generic 5mm and 3mm through-hole LEDs.

These are very common parts, and we should have generic packages for them.

Generator PR: https://github.com/LibrePCB/librepcb-parts-generator/pull/91

Note: I did not add generic 8mm and 10mm LEDs. These are much less common, and differences in outer diameter may make a bigger difference in absolute numbers.

#### SCREENSHOTS

5mm

screenshot-20220226-150248

3mm

screenshot-20220226-150257

##### DATASHEETS / REFERENCES

Because there's no standard, I looked at the 500 most popular THT 5mm LEDs on Digikey. Here's the height distribution:

screenshot-20220225-223813

From these values, it seems that 8.9mm, 10.3mm and 13.3mm seem to be common heights, so I picked those three.

The main diameter is fixed at 5mm, but the "outer" diameter of the bottom ring differs between manufacturers. I sorted the list by number of pieces in stock (most popuplar parts will have larger stock) and picked three top models from different manufacturers.

- Cree C503B-RAN-CA0B0AA1
  -> Ring diameter: 5.6-5.8mm
- Liteon LTL-307Y
  -> Ring diameter: 5.85mm
 - Seoul LW514
  -> Ring diameter: 5.6

Based on these values, I piked "5.8mm" as ring diameter, which seems reasonable. Minor deviations from this should not be an issue.

I did the same thing for 3mm LEDs.

- [ ] Regarding the color: I only generated the package for a clear white LED. Others could easily be generated as well, but it might result in an explosion of variants. Not sure if we want that or not.
- [x] Should a different category be used? (Yes, Urban will create one.)
- [x] Once initial feedback is here, I want to generate 3mm LEDs as well.
- [x] Use hyphens for name
- [x] Rectangular cathode pad
- [x] Add "Generated with librepcb-parts-generator" to description

##### CHECKLIST

  - No IPC names defined for THT LEDs


dbrgn:
Updated, 3mm LEDs have been added as well.

ubruhin:
> Regarding the color: I only generated the package for a clear white LED. Others could easily be generated as well, but it might result in an explosion of variants. Not sure if we want that or not.

With "white" you mean the color of the light? IMHO the color of the light is not relevant for the package, but only the color of the lens. Thus I'd declare it simply as "clear".

> Should a different category be used?

Since we already have a category "Diode -> Chip LED" I'd say it makes sense to add a "Diode -> Through-Hole LED" as well. Should I add it?

Regarding the footprints:

- For all the other non-IPC package names we use hyphens as separator ("R-THT-0207", "DIP-8", "C-0603", ...), therefore I'd prefer dashes in the LED packages as well to keep it consistent. As far as I can see in the IPC naming convention, it is totally fine to use hyphens here.
- Maybe make the cathode pad rectangular to see the polarity even on PCBs without silkscreen? At least this seems to be quite common, for example the KiCad LEDs also do that. Octagon might also make sense since it is more space efficient, although this would be rather unusual.
- I think for the 5mm LEDs the same pad geometry could be used as for the 3mm LED, i.e. circular/square instead of an oblong (oblongs are wasting quite a lot of PCB space...)
- Usually the generated parts contain "Generated with librepcb-parts-generator" in their description, but these LEDs don't.
- Actually LEDs would be perfect candidates for adding additional footprints for horizontal mounting since that's a typical use-case (placed at the PCB edge 90° bend through a front panel) :wink: Shall I try to add this to the generator?

dbrgn:
> With "white" you mean the color of the light? IMHO the color of the light is not relevant for the package, but only the color of the lens. Thus I'd declare it simply as "clear".

A clear white LED and a diffused white LED will look different. The naming is here for consistency reasons, this way we can add a "Clear Red" or a "Diffused Green" LED as well.

> Since we already have a category "Diode -> Chip LED" I'd say it makes sense to add a "Diode -> Through-Hole LED" as well. Should I add it?

Yes please 🙂

> I think for the 5mm LEDs the same pad geometry could be used as for the 3mm LED, i.e. circular/square instead of an oblong (oblongs are wasting quite a lot of PCB space...)

The oblongs are easier to solder though... On a PCB without space constraints, I'd usually prefer oblongs. And you won't be able to place other components below the LED anyways... (Possibly on the bottom side, but if you have such space constraints you most probably won't be using 5mm through-hole LEDs.

> Actually LEDs would be perfect candidates for adding additional footprints for horizontal mounting since that's a typical use-case (placed at the PCB edge 90° bend through a front panel) wink Shall I try to add this to the generator?

Hm, is it common to take regular LEDs and bend their legs 90°? Wouldn't you usually just order LEDs with angled legs, or specialized angled LEDs (like [ŧhis](https://www.ledtronics.com/Products/product_new.aspx?id=PCT300TWRG4-3A) or [this](https://www.digikey.ch/de/products/detail/vishay-semiconductor-opto-division/TLPR5600/4074661))? In that case, if the devices are different, there should be different packages as well (since you cannot hardcode a package variant for a device).

I'll apply the other suggestions.

dbrgn:
All suggestions implemented (except for the horizontal footprint variants and smaller pads).

dbrgn:
Package category updated.

ubruhin:
> A clear white LED and a diffused white LED will look different. The naming is here for consistency reasons, this way we can add a "Clear Red" or a "Diffused Green" LED as well.

Hmm OK but then why adding a *white* LED as a generic LED, instead of a *transparent* one? LEDs with transparent body are much more common than LEDs with white body. In addition, LEDs with transparent body are available with all the different light colors, so such a package would fit even for light colors other than just white. Thus IMHO it makes much more sense to make the generic LED transparent ;)

And I'm not sure if we should really distinguish between clear and diffused body. In the end the difference is very small and even once we support 3D models, with STEP you probably still can't distinguish between these two surface properties. So to be honest, I don't think we will ever generate even more generic LED packages just to have separate packages for clear and diffused surface. I mean, we are talking only about *generic* LEDs here - when creating packages for specific parts (with manufacturer-provided land pattern and 3D model) these package will be named according the part name anyway, and it's not relevant at all how the generic LED packages look.

Does this make sense, or may I still don't understand exactly what you mean with "clear white"? :see_no_evil: 

> The oblongs are easier to solder though... On a PCB without space constraints, I'd usually prefer oblongs. And you won't be able to place other components below the LED anyways... (Possibly on the bottom side, but if you have such space constraints you most probably won't be using 5mm through-hole LEDs.

Of course the space is very relevant - you can place parts on the bottom of the LED, and even much more important, you can route traces *on any layer* (top, inner, bottom) below the LED. Which doesn't work if the pads are too large, and the only option would be to create your own package with smaller pads. The other way around is much less cumbersome - you have no space issues, just soldering is *very slightly* more difficult, but still easy to handle - so you can get your project done.

An alternative would be to create two separate footprints, one with round pads and one with oblongs. What do you think?

> Hm, is it common to take regular LEDs and bend their legs 90°? Wouldn't you usually just order LEDs with angled legs, or specialized angled LEDs (like [ŧhis](https://www.ledtronics.com/Products/product_new.aspx?id=PCT300TWRG4-3A) or [this](https://www.digikey.ch/de/products/detail/vishay-semiconductor-opto-division/TLPR5600/4074661))? In that case, if the devices are different, there should be different packages as well (since you cannot hardcode a package variant for a device).

Of course at least for hobby usage it is very common to use standard LEDs (which you might even have in stock) for everything and then just bend them as you need it. And keep in mind that we are still talking only about *generic* LED packages - these are intended for exactly the use-case where you don't want to create part-specific packages. For professional usage, you would create part-specific packages anyway where I agree it doesn't make sense to provide different footprints for the different mounting variants.

However, at least we could add the additional footprints some day later...

dbrgn:
Hmm, my thinking was that we want a light color specific package (i.e. blue light, red light, white light), and that there are both clear and diffuse bodies. That's two different dimensions (light color and package color). The reason was that you might want to add both a red and a green LED in your 3D model, and those are usually diffused, but if you add a white LED it's usually clear. Thus, I added both these "dimensions" to the name for consistency.

If we only create a single generic LED package, we can just make it clear without specifying a light color (most clear LEDs have a white/transparent package, although I've seen red or green LEDs in the past with a clear but tinted package). I think that actually makes sense, instead of adding all possible color variants, even when creating custom components of colored LEDs, most users probably don't care about the color of the 3D model. Thus, if you agree, I'll rename the package to something like "LED-THT-D500H1030-CLEAR".

> Of course the space is very relevant - you can place parts on the bottom of the LED, and even much more important, you can route traces on any layer (top, inner, bottom) below the LED. Which doesn't work if the pads are too large, and the only option would be to create your own package with smaller pads. The other way around is much less cumbersome - you have no space issues, just soldering is very slightly more difficult, but still easy to handle - so you can get your project done.

Tell me, which project in 2022 would use a multi-layer PCB, and have such tight board space constraints that you would place components and traces below the LED, but *still* use a gigantic 5mm through-hole LED? 🙈 To me, a 5mm through-hole LED is one of the components that is almost purely used in the hobbyist domain (or in designs from the 90s), for other use cases you'd use SMD parts. And in the hobbyist domain, solderability is much more important than the board space below an LED footprint.

If you insist that this is useful, I can add two footprint variants (after all it's quite easy to do), although I think it's a waste of time.

> Of course at least for hobby usage it is very common to use standard LEDs (which you might even have in stock) for everything and then just bend them as you need it.

Ok, I think that's a valid use case, but then we might need multiple footprint variants with different offsets from the holes, similar to the through-hole resistor footprints that have 3 different hole distance variants, right?

These footprint variants should definitely be added in a separate PR though.

dbrgn:
@ubruhin Alright, PR is updated, I dropped the "white" part (see https://github.com/LibrePCB/librepcb-parts-generator/pull/91/commits/8a7c9e452396cd0b0a9434d2021284fe70177622).

I'd suggest adding footprint variants in future commits, in order to merge this soon.

dbrgn:
@ubruhin reminder 🙂

ubruhin:
> However, it makes sense to drop the light color dimension, since I don't expect 3D models to be that detailed anyways. If we only want a single generic LED package, we can just make it clear without specifying a light color (most clear LEDs have a white/transparent package, although I've seen red or green LEDs in the past with a clear but tinted package). Even when creating project-custom devices for colored LEDs, most users probably don't care about the color of the 3D model and would be happy with a generic clear one. Thus, if you agree, I'll rename the package to something like "LED-THT-D500H1030-CLEAR".

Thanks for repeating what I wrote in my very first comment https://github.com/LibrePCB-Libraries/LibrePCB_Base.lplib/pull/92#issuecomment-1052793518, before starting a long discussion about the color :joy: So yes, I agree!

> Tell me, which project in 2022 would use a multi-layer PCB, and have such tight board space constraints that you would place components and traces below the LED, but still use a gigantic 5mm through-hole LED?

Even in 2022 it's often simpler to use a THT LED instead of building something complex with lightpipes to forward the light into a frontpanel - and space is almost always a problem, no matter whether professional or hobbyist PCB :wink: So I'm really sorry but I'd indeed prefer to use round/square pads :see_no_evil: Note that both Eagle and KiCad use small pads as well, even up to much larger LEDs :nerd_face:

dbrgn:
@ubruhin do you agree with this?

> I'd suggest adding footprint variants in future commits, in order to merge this soon.

If yes, is there anything else that blocks this from getting merged?

ubruhin:
Updated the generated files with the changes from this new generator PR: https://github.com/LibrePCB/librepcb-parts-generator/pull/97

dbrgn:
@ubruhin looks good to me! 👍🏻 The footprint variants are nice.

PR #94, opened 2022-05-08
Add devices for SOD{1,3,5}23 diodes

Tubbles:
##### SUMMARY

Added devices for SOD123, SOD323, and SOD523 diodes
The packages already existed (created by Danilo B.), so I mapped them to devices.

##### DATASHEETS / REFERENCES

Generic devices

##### CHECKLIST


dbrgn:
Hmm, I guess the question is how much we want to have generic devices 🙂 Your call, @ubruhin.

ubruhin:
> Hmm, I guess the question is how much we want to have generic devices

Yeah that's a good question. I understand that such devices are indeed useful in some cases, but it feels wrong to me because there's no clear standard about the pinout of generic devices. Actually you don't know which real device you can solder on the PCB if you used a generic device, so this is kind of error-prone.

At the moment I prefer not adding generic devices to the official libraries. I suggest to create concrete devices instead (especially since it takes only a minute), or of course feel free to add generic devices to your personal library.

So I'll close this for now - sorry, and thanks anyway.

dbrgn:
> I recommend to create concrete devices instead (especially since it takes only a minute)

@Tubbles it would be nice if you could contribute your concrete diodes to the library 🙂 Since component and package already exist, it should be very quick to add them.

PR #99, opened 2023-07-06
Add sym/cmp for polyfuses

ii8:
Add Polyfuse symbol and component

##### CHECKLIST

One thing I should add: the polyfuse pin numbers appear on the schematic editor, but they don't show for regular fuses, I don't know why that is or how to change it. It would be better if they did not appear for polyfuses either I think.


ubruhin:
Thanks! I allowed myself to make some modifications, I hope that's fine for you.

- Remove line through the symbol body (AFAIK polyfuses usually don't contain that line, please correct me if I'm wrong)
- Add US version of symbol
- Remove category, moving to "Resistors" category instead (A category "Fuses" within the "Passive" category somehow feels wrong, as most fuses are not passive components - but polyfuses are actually resistors, so using that category should be fine)
- Hide symbol pin names in schematics

image

ii8:
Of course that's fine.
I saw in KiCad they have a line through it but you might be right that it's more common to not put the line.
Looks good to me ty.

PR #100, opened 2023-07-24
Add sym/cmp for crystals with a grounded case

ii8:
##### SUMMARY
add crystals with ground connection

##### CHECKLIST


ubruhin:
@ii8 Thanks for this contribution and sorry for the delay! Just one quick question: Do you have any objections in making the symbol more compact and keeping the same orientation as the other crystal symbol?

image

(Maybe some day we should rotate them both to have GND at bottom, but the project library updater is not able yet to auto-correct the rotation after an update so it's better to keep the rotation of the existing crystal for now.)

If you're fine with this change, I'd commit it and merge this.

ii8:
Yea that's fine

PR #109, opened 2023-09-06
Add sym/cmp for HD44780 compatible displays

ubruhin:
##### SUMMARY

There are many displays with a HD44780 compatible interface so they can all be represented by the same symbol and component. Just the backlight may or may not be available, thus added both variants.

With backlight:
image

Without backlight: 
image

##### DATASHEETS / REFERENCES

- Wikipedia: https://en.wikipedia.org/wiki/Hitachi_HD44780_LCD_controller
- Example datasheet: https://newhavendisplay.com/content/specs/NHD-C0216AZ-FSW-GBW.pdf

None.

##### CHECKLIST


dbrgn:
First of all, does this really belong in the base library? Shouldn't it be part of `LibrePCB_ICs.lplib`?

Nice, that you marked pins 0-3 as optional and 4-7 as required! Most times, these displays are run in 4-bit mode.

Some keywords would be nice. Maybe "hd44780,hitachi,lcd,character lcd"?

The name "V0" seems a bit non-intuitive for the contrast pin. Wikipedia calls it "Vee", which isn't better. On all my LCD modules I have laying around (5, although I should have some more, which I can't find right now), if the pin is labelled on the silkscreen, it _is_ named "V0". I wonder if "Contrast" might be better? (However, if "V0" is the established term, it's probably best to stick to it.)

You named the backlight pins "LED". Was that on purpose? To quote WP:

> Character LCDs may have a [backlight](https://en.wikipedia.org/wiki/Backlight), which may be [LED](https://en.wikipedia.org/wiki/Light-emitting_diode), [fluorescent](https://en.wikipedia.org/wiki/Fluorescent_lamp), or [electroluminescent](https://en.wikipedia.org/wiki/Electroluminescence). The nominal operating voltage for LED backlights is 5V at full brightness, with dimming at lower voltages dependent on the details such as LED color. Non-LED backlights often require higher voltages.

(Changing this would also have an effect on the descriptions.)

ubruhin:
Thanks for taking a look at this!

> First of all, does this really belong in the base library? Shouldn't it be part of LibrePCB_ICs.lplib?

First I had exactly the same thought! :smiley: Mainly because HD44780 *is* an IC. But this sym/cmp does not represent the IC (it has many more pins) but an LCD module which is *not* an IC. I mean, it doesn't fit perfect for either library but somehow the base library feels more intuitive for me as most of the other optoelectronic parts are here as well.

> Some keywords would be nice. Maybe "hd44780,hitachi,lcd,character lcd"?

Will do! Maybe just without hd44780, actually I think we're overusing keywords a little bit :joy: Every term already contained in the name does not need to be duplicated in the keywords.

> The name "V0" seems a bit non-intuitive for the contrast pin. [...]

I agree it's not intuitive. But I also saw that this seems to be the most widely used term so I think it makes more sense than something like "contrast".

> You named the backlight pins "LED". Was that on purpose?

To be honest, so far I'm not aware of a display calling it other than LED+/-. But probably there exist other backlight types. I just wonder whether they should have a separate component then :thinking: Most probably the driver circuit of the backlight needs to be quite different for different kinds of backlights... For example naming it BL_VCC and BL_GND (or something like that) could be misleading since not a specific *voltage* is required, but a specific *current*.

dbrgn:
> To be honest, so far I'm not aware of a display calling it other than LED+/-.

All my labelled displays call it A/K 🙂

dbrgn:
> Every term already contained in the name does not need to be duplicated in the keywords.

Yeah, I wasn't sure about that, so I put it in the list again 🙂 Makes sense though.

> To be honest, so far I'm not aware of a display calling it other than LED+/-.

All my labelled displays call it A/K 🙂 Which isn't really better, in my opinion.

You're probably right. LED is by far the most common backlight technology, and we can just assume it's using LED.

Maybe we should rename the part to "with LED Backlight", but that's probably overkill.

ubruhin:
> All my labelled displays call it A/K 🙂 Which isn't really better, in my opinion.

Hmm okay, I agree it's not better :see_no_evil: So let's keep LED+/- for now.

> Maybe we should rename the part to "with LED Backlight", but that's probably overkill.

I added it, so now the scope of the sym/cmp should be very clear ;)

Keywords added as well.

dbrgn (approved):
LGTM! (But the fixup went wrong.)

PR #123, opened 2023-09-27
Add sym/cmp for bridge rectifier

Recycled0080:
##### SUMMARY

added sym/cmp 
Screenshot from 2023-09-28 10-04-34
for bridge rectifier both EU and US

##### CHECKLIST

- [ x] I have read and followed the library conventions¹.
- [ ] For modified elements, I bumped their version number².
- [ x] I already used the new library elements in a design and verified their correctness (optional).
- [x ] I'm the copyright owner of the added content (i.e. the changes are made by myself, not copied/imported from somewhere else).
- [ x] I agree to publish all my changes under the CC0 Public Domain License³, allowing everyone to use and modify the content without any restrictions.


ubruhin:
Very nice!

Just two ideas:

- Could the symbols be slightly shrinked to get the pins located on a 4x4 grid instead of 6x6? I think it would help to keep schematics compact. Note that the pin length could be shortened a bit (e.g. to 1.27mm) for this purpose.
- Probably it would look better to enable the grab area on the diode triangles, for consistency with single diodes?

Recycled0080:
I have shrunk the symbols down to 4x4, I added extra grab areas to triangles, picture below,
Screenshot from 2023-10-02 12-41-38
Screenshot from 2023-10-02 11-54-29

ubruhin:
The 4x4 looks great IMO :+1: Regarding the grab area, maybe I was not clear enough, sorry. I meant to enable the "Grab Area" property of the already existing triangle polygons so the diodes are filled with yellow color the same way as single diodes. I quickly changed this by myself now (and also moved text labels in the EU symbol to the same position as in the US symbol):

image

Thanks for this contribution! :tada:

PR #126, opened 2023-10-06
Add dev for standard THT transistors BC5xx

ubruhin:
##### SUMMARY

Adding the good old THT transistor devices from the BC5xx family, including MPNs:

image

##### DATASHEETS / REFERENCES

- https://en.wikipedia.org/wiki/BC548
- https://en.wikipedia.org/wiki/Pro_Electron
- https://digikey.com

Many many questions :see_no_evil: 

1. Normally we do not add devices of specific parts to the base library since they belong to the manufacturer-specific library. However, the BC5xx transistors might be considered as generic enough to add them as devices here (Wikipedia even says those are somehow standardized, see links above). Especially from the user point of view, I'd expect those transistors to be available when installing just the base library (what could be considered as _base_ if not those transistors?). Or in other words, it would be a bit surprising to install the base library and still don't have such basic transistors available.
2. Given we agree on 1. to add these devices to the base library, the question arises whether we should actually add MPNs to them or not. Somehow it feels unnecessary (those transistors are mostly used by hobbyists having them in stock and probably not needing an accurate BOM; in addition, it turned out that tons of MPNs are considered as obsolete, not sure how long those will still be sold). On the other hand, adding them shouldn't hurt (except that it is time consuming).
3. Should we really add *all* those transistors? At least a few of them are not even available on Digikey anymore (but still listed as "active" on Octopart).
4. Most of these transistors are sold in a 1.27mm pitch TO-92 as well as in a 2.54mm pitch TO-92 (bent legs). Thus I added two devices for each transistor type, the 1.27mm variant called "bulk" since according datasheets all those transistors are packaged as bulk instead of tape&reel/ammo. Here also the question arises whether this is overengineered or not?
5. One might even consider adding just 4 devices in total, called "BC546..550 [(bulk)]" (NPN) and "BC556..560 [(bulk)]" (PNP) and distinguish the exact type just by the MPNs. However, I think having explicit devices is useful for hobbyists since it allows just to add a "BC548" without requiring to choose an exact MPN.

A completely different approach would be to add only very generic transistors to the base library (without MPNs, only one pitch variant) and add the same transistors to the manufacturer-specific libraries as well, with MPNs etc.

Maybe @dbrgn has any opinion about that? :slightly_smiling_face: 

Btw, the same questions arise for other very generic devices, for example an 1N4007 diode I'd expect to be available after installing the base library. Probably there exist many more such devices (Zener diodes, NE555, ...) which makes it hard to decide where to draw the line between generic devices and specific devices :sob: Generally it might still be a solution to decide against this convenience, requiring people to install manufacturer-specific libraries for such devices.

##### CHECKLIST


ubruhin:
Considering how many "complaints" we get about missing basic library elements, I think we *should* do add these generic transistors (including MPNs, as they do provide a benefit). Adding them doesn't hurt anyway, if we ever decide against such generic devices we can safely remove them as nothing breaks when deleting devices.

Though probably I wouldn't go too far with adding generic devices. Let's add things like 1N4007 as well (the really basic things for hobbyists), maybe also generic OpAmps but not more specialized ICs like NE555.

dbrgn:
Both options (this PR and fully generic devices) seem valid choices to me. If you want to discuss it in more detail, we could do that at Coredump sometime.

PR #138, opened 2024-04-25
Add more attributes to capacitor components

### LINKED ISSUE #137: Add voltage rating value in the symbol of capacitor
Maybe it will be nice to have in the capacitor symbol not only the display of the capacitance value as VALUE but also the rating voltage value.

ubruhin:
##### SUMMARY

- Add attributes `RATED_VOLTAGE` (unipolar + bipolar) and `DIELECTRIC` (bipolar only).
- Also add these attributes to the component's default value to make them appear in schematics

Closes #137

##### DATASHEETS / REFERENCES

None

The schematic editor does not handle the new default value very well when adding new capacitors to the schematic, thus we should not merge this yet...

##### CHECKLIST


eduardosm:
> The schematic editor does not handle the new default value very well when adding new capacitors to the schematic, thus we should not merge this yet...

Has this been fixed by https://github.com/LibrePCB/LibrePCB/pull/1807?

ubruhin:
> Has this been fixed by https://github.com/LibrePCB/LibrePCB/pull/1807?

I think so, thus I just updated this PR yesterday. I'll need to test, review and think about this again to decide if it is now ready to merge.

Also I think we should still wait a bit with merging this. There are still many people using LibrePCB 2.0. With this version, it would be bad to get the updated capacitors as the editor won't behave nicely with it. It's not a big deal for less used components, but the capacitor is one of the most used component so it would be very bad user experience.

[Packaging status](https://repology.org/project/librepcb/versions)

PR #148, opened 2024-11-21
Relay coil/solenoid symbol - change pin text alignment

DoppioT:
##### SUMMARY
Change the alignment of the text on pins of a Relay Coil/Solenoid symbol

When the relay coil is used in a 2-coil relay, in which the pins are called Set+/- Rst+/-, the text of the two pins overlaps
image

This PR changes the pin text offset from left-center to center-bottom
image

This of course also impacts the 1-coil relay,  but I'd say that the component remains readable, IMHO
Before:
image
After:
image

##### DATASHEETS / REFERENCES
N/A

In the 2-coil relay, the pin text now slightly overlap with the border. It seems still preferable to left-center alignment which causes the + and - signs in the 1-coil relay to be far from their pin
image

##### CHECKLIST

- [X] I have read and followed the library conventions¹.
- [ ] For packages, I followed IPC7351C (see details in library conventions). (didn't change)
- [X] For modified elements, I bumped their version number². 
- [X] I'm the copyright owner of the added content (i.e. the changes are made by myself, not copied/imported from somewhere else).
- [X] I agree to publish all my changes under the CC0 Public Domain License³, allowing everyone to use and modify the content without any restrictions.


ubruhin:
Hi,
thanks for pointing out this problem.
Wouldn't it look even better if rotating the texts by 90° so they can stay centered?

DoppioT:
>  Wouldn't it look even better if rotating the texts by 90° so they can stay centered?

You're right 
image
image

PR updated

ubruhin:
Perfect, thank you! :+1:

PR #161, opened 2025-10-04
Added BC817/BC807

eto-:
##### SUMMARY
Added BC817/BC807 complementary sot23 transistors

##### DATASHEETS / REFERENCES
https://www.mouser.it/datasheet/3/175/1/BC807-16_-25_-40.pdf
https://www.mouser.it/datasheet/3/175/1/ds11107.pdf

I added the models available on Mouser for the most common producers.
If these devices are fine, I can add others sot23 transistors.

##### CHECKLIST

- [ ] For modified elements, I bumped their version number².


ubruhin:
Great choice to make a small PR, not all at once :smiley: 

I didn't review the MPNs exactly yet. But I have some feedback for the rest:

- The PNP devices uses the wrong component (NPN instead of PNP).
- The package looks wrong. According your linked datasheets, it should be "SOT23-3P95_230X110L50X40" (always be careful between "SOT23" and "SOT23-3" which are not the same).
- The manufacturer names are not consistent with other parts. They should use original capitalization as used on their own websites, Digikey etc, respectively consistent with the existing parts in our libraries. Those are:
  - "onsemi" (that one is fine as-is)
  - "diodes inc." -> "Diodes Incorporated"
  - "nexperia" -> "Nexperia"
- I think it would be nice to add a datasheet URL to the devices. Even though they wouldn't be valid for all the different MPNs, I'd say in this case we can allow that since those devices are considered generic anyway. In future we can even add multiple datasheets for the different manufacturers, it's just not yet supported by the UI.

eto-:
> * The PNP device uses the wrong component (NPN instead of PNP).

I was sure I changed that ... fixed

> * The package looks wrong. According your linked datasheets, it should be "SOT23-3P95_230X110L50X40" (always be careful between "SOT23" and "SOT23-3" which are not the same).

Yes, I'm not able to select it right now:
image

Moreover there is confusion in the keywords: for SOT23 (the smaller) there is a "sot23-3" keyword and for the SOT23-3 (the bigger) there is a "sot-23" keyword. 

> * The manufacturer names are not consistent with other parts. They should use original capitalization as used on their own websites, Digikey etc, respectively consistent with the existing parts in our libraries. Those are:

Fixed

ubruhin:
> Yes, I'm not able to select it right now:

Hmm that is strange. But you see both SOT23 packages in the library?

> Moreover there is confusion in the keywords: for SOT23 (the smaller) there is a "sot23-3" keyword and for the SOT23-3 (the bigger) there is a "sot-23" keyword.

That is by intention. The terms "SOT23" and "SOT23-3" are sometimes not used correctly (IIRC even in datasheets it is sometimes misleading). By having both keywords in both packages, the user will always find both of them when searching for sot23, so he has to explicitly choose the correct one. If we didn't have those keywords, you can search for "sot23-3" and you will just take the single package which was found, even though it might be the wrong one.

ubruhin:
I have no idea why you don't see both SOT23 packages, while both are visible in the library. I just checked and for me it is there:

image

Would be interesting to understand how this can happen. If you are willing to investigate:
- Maybe there's any interesting logging output when running librepcb from the terminal
- Maybe the base library (or specifically that package) exists multiple times in your workspace (e.g. as a local library and as a remote library). And maybe one of those don't have the "SOT" category assigned for some reason, which would explain that behavior.

eto-:
I found that there is some sort of conflict with a local library containing devices not yet committed.
I will try to debug later. 

In the meantime I fixed the packages for the two transistors.

ubruhin:
Alright, I checked the MPNs, did some reordering and added a few more parts to BC807. I think now it looks good to me.

eto-:
> * Maybe the base library (or specifically that package) exists multiple times in your workspace (e.g. as a local library and as a remote library). And maybe one of those don't have the "SOT" category assigned for some reason, which would explain that behavior.

Indeed there was another package with the same UUID and funny it was a [sot variant](https://www.infineon.com/assets/row/public/documents/24/49/infineon-bga614-ds-en.pdf?fileId=db3a304314dca3890115418ffd35163a)

It was done when I was fresh new on librecad and probably I made a mistake.

I did not see any warning or error at the beginning: maybe we should add a warning if there are replicated UUID. I can write a small script to search replicated UUID if needed.

ubruhin:
Ok thank you for checking, then it makes sense this happened.

> I did not see any warning or error at the beginning: maybe we should add a warning if there are replicated UUID.

The library editor does show a big warning when you change the package or component of a device. But after it has been ignored once, there is no more warning afterwards. It would indeed be nice to detect & report this kind of problem, I opened https://github.com/LibrePCB/LibrePCB/issues/1620 to keep this in mind.

PR #193, opened 2026-04-26
Add symbols and components for 2/3/4-line common mode chokes

eduardosm:
##### SUMMARY

Add symbols and components for common mode chokes with 2/3/4 lines

##### CHECKLIST

- [ ] ~For packages, I followed IPC7351C (see details in library conventions).~
- [ ] ~For modified elements, I bumped their version number².~


ubruhin:
That is a difficult one :see_no_evil: 

Symbols:
- It seems that usually the pairs of lines are visually indicated by having the straight lines *between* the coils. I looked at several datasheets, searched in the internet, and checked also the KiCad libraries and it almost always looks like this:
  image
  image 
  Only the 3-line choke is usually drawn like yours.
- Maybe the vertical pin spacing should be 5.08mm - at least for the symmetric variants there will be more space needed anyway. I think the 3-lines choke can keep 2.54mm spacing.
- In the schematic, this looks like separate inductors. There's no clear visual difference between two single inductors and a 2-line choke. Since those chokes are all within a single package, somehow I think this should be expressed by drawing them inside of a yellow polygon (like ICs). Then it will be clear that this is a single device. Or would this look bad in the schematic? :thinking: 
- Not sure if the names should be "Choke Common Mode" instead of "Common Mode Choke", according our [library conventions](https://librepcb.org/docs/library-conventions/symbols/#naming) to have the generic term in the front. But no clear opinion, I think I'm fine with both variants.

Components:
- Just a detail: I saw KiCad uses "FL" as prefix for common mode chokes. Not sure where this is coming from and if some norm specifies a prefix. In the internet I see "L" more than "FL". No clear opinion from myself, I'd say "L" is fine unless a norm explicitly specifies something else.
- My biggest question mark here is the default value. I think in many cases, the MPN is the most important attribute of a common mode choke and should be visible in the schematic, therefore `{{MPN}}` or maybe better `{{MPN or DEVICE}}` should be part of the default value. And for the electric attributes, it seems that depending on the use-case, one or another attribute is important. Sometimes inductance, but I think very often the inductance is not even known, as only inductance (e.g. @100MHz), DC resistance, and rated current are of interest. So it's really hard to define a reasonable default value. Maybe something like this?
  ```
  {{INDUCTANCE}} {{IMPEDANCE}} {{DC_RESISTANCE}} {{RATED_CURRENT}}
  {{MPN or DEVICE}}
  ```
  Depending on which attributes you fill out, it could then evaluate to something like this:
  ```
  20μH   0.16Ω 500mA
  WE-SL3
  ```
  or this
  ```
    30Ω@100MHz 0.2Ω 400mA
  SRF2012-300YA
  ```
  Though the whitespaces are very ugly and also it's a bit cumbersome in general to have so many attributes. I think for long term we should make LibrePCB stripping unnecessary whitespaces from the evaluated VALUE to at least fix this problem.

ubruhin:
I implemented https://github.com/LibrePCB/LibrePCB/pull/1807 to hopefully get rid of the whitespaces in attribute substitutions starting with LibrePCB 2.1.0.

eduardosm:
> It seems that usually the pairs of lines are visually indicated by having the straight lines _between_ the coils. I looked at several datasheets, searched in the internet, and checked also the KiCad libraries and it almost always looks like this:

Changed the 2- and 4-line versions, so it looks like this:

2-line
4-line


> In the schematic, this looks like separate inductors. There's no clear visual difference between two single inductors and a 2-line choke. Since those chokes are all within a single package, somehow I think this should be expressed by drawing them inside of a yellow polygon (like ICs). Then it will be clear that this is a single device. Or would this look bad in the schematic? 🤔

I don't think the 2-line choke looks like two inductors. The core bars and polarity mark should hint that the inductors are couple together.

However, the 4-line choke now looks like two 2-line chokes. This is how it looks boxed:

4-line-boxed


> My biggest question mark here is the default value. I think in many cases, the MPN is the most important attribute of a common mode choke and should be visible in the schematic, therefore `{{MPN}}` or maybe better `{{MPN or DEVICE}}` should be part of the default value. And for the electric attributes, it seems that depending on the use-case, one or another attribute is important. Sometimes inductance, but I think very often the inductance is not even known, as only inductance (e.g. @100MHz), DC resistance, and rated current are of interest. So it's really hard to define a reasonable default value. Maybe something like this?

Should the impedance value be a string (so it can include the "@frequency")? Or be two values, one for the actual impedance and one for the frequency. The second option would leave a floating "@" if values are not filled. Regarding DC resistance and rated current, those are typical parameters of regular inductors too, but they are not usually shown on schematics.

ubruhin:
Sorry for the late response, was quite busy with the 2.1.0 release. Now back to this topic to hopefully get it merged soon.

> Changed the 2- and 4-line versions, so it looks like this:

Looks good to me :+1: 

> However, the 4-line choke now looks like two 2-line chokes.

Hmm in the datasheets I found, it looks exactly like this. Like two pairs of lines. Is there even a coupling between all 4 lines / do you have a concrete datasheet which does have it? Maybe actually both variants exist, devices with two independent pairs of lines, and devices with 4 coupled lines.

> This is how it looks boxed:

I like this look. IMHO, it makes clear that this is a single device. For me, it would feel inconsistent to the rest of the library without the box.

> Should the impedance value be a string

Yes I'd define it as a string to avoid leaving behind a standalone '@'.

> Regarding DC resistance and rated current, those are typical parameters of regular inductors too, but they are not usually shown on schematics.

Yeah, but maybe it would still be good to include them in the default value. It is easier to remove those in the schematic if not desired, rather than adding more values if they are desired. By default, if the actual device doesn't define those attributes, they won't appear in the schematics anyway, so it doesn't hurt to have the (unused) placeholders in the default value. LibrePCB 2.1.0 can now handle this situation nicely as it trims whitespaces from the value.

I think my suggestion would now be this default value:

```
{{MPN or DEVICE}}
{{INDUCTANCE}} {{IMPEDANCE}} {{DC_RESISTANCE}} {{RATED_CURRENT}}
```

(I moved MPN/DEVICE to the beginning, as I think this is the most important attribute, and also the only one that is guaranteed to be non-empty - all the other attributes may be undefined).

Would you agree to this? Or what would be your suggestion?

eduardosm:
Boxed the 4-line symbol and changed the default value to your suggestion.


For the 4-line here is a datasheet: https://yageogroup.com/content/datasheet/asset/file/KEM_LF0073_M-500CT
Which has the two variants (all four coupled and coupled in pairs):

datasheet

This is how the "Z" symbol looks:

4-line-z

So maybe we should have a "Common Mode Choke 4 Lines" and a "Common Mode Choke 2 Lines Dual"?

ubruhin:
Alright, so both variants exist. I'd say the two pairs are quite generic, but about the coupled 4-line variant I'm not sure if that is generic or very specific to the posted device. I mean, the "Z" goes from coil 2 to coil 4. Does this have any meaning? May there also exist other "Z" variations, like from coil 2 to 3, 1 to 3, etc? Or does the "Z" just mean "all lines are coupled together"? I have no idea.

At the moment, honestly my suggestion would be to add only the generic (i.e. 2 pairs) to the base library. About the other one we may discuss another time - or is this exactly the variant you actually need now?

Did you intentionally box only the 4-line symbol, but not the other symbols? Actually I was thinking about boxing all symbols. IMHO it would look better, and also more consistent.

Regarding naming: Yeah, probably it would be good to rename "4 Lines" to "2x2 Lines" to make the coupling clear.

Except from those two things, I think everything looks good to me.

eduardosm:
Ok, renaming "4 Lines" to "2x2 Lines" for now.

The "Z" means all inductors crossed by the line are coupled.

I'm not sure how to box this case:

If I leave 2.54mm between pin and top, the symbol looks too close to the edge:
a

If I increase marging by 1.27mm, margins look good, but there isn't 2.54mm between pin and top anymore:

b

Or leaving 5.04 between pin and top and between pin and bottom:

c

ubruhin:
I'd go with the second variant, IMO the asymmetric top/bottom distances don't look bad.

Btw, I just realized that the coils are quite some bigger than the standard inductor symbol. When using the same size as the standard inductor, I'd say the symbol could be made 5.08mm less wide (2.54mm on each side):

image

Though it's not a must for now. I already told you I'm fine with the look :slightly_smiling_face: I could still try by myself later if the smaller variant really looks better and is worth to adjust.

eduardosm:
I had copied the coil from the transformer symbol.

This is how it looks:

image

I you greenlight it, I'll change the others.

ubruhin:
> I had copied the coil from the transformer symbol.

Ah, I see.

I'd say the small variant looks good and is space efficient -> green light :slightly_smiling_face:

eduardosm:
Done 🙂

image

ubruhin (approved):
Perfect 👌 

Thanks for your patience with my picky feedback! 🙂

